PR #6, opened 2018-05-02
Use cryptography library to load attestation cert instead of pyOpenSSL

dsanders11:
It's simpler to read the attestation cert with `cryptography`, and moves a step closer toward being able to drop `pyOpenSSL` as a dependency (using two different crypto libraries is kinda icky).


futureimperfect:
Looks good to me, thanks!

PR #7, opened 2018-05-02
Fix conflicting versions of cryptography and pyOpenSSL

dsanders11:
There is a conflict with the versions of `pyOpenSSL` and `cryptography` which can break `pip` in the Docker image amongst other problems.


futureimperfect:
Looks good to me, thanks!

PR #8, opened 2018-05-02
Switch demo to using HTTPS as it is required for Chrome usage

dsanders11:
Chrome will return `undefined` for `navigator.credentials` unless the site is using HTTPS, rendering the current demo unusable. Luckily Flask has built-in support for serving over HTTPS with a self-signed certificate.

This is a resubmission of #5, which was accidentally not on its own branch.


futureimperfect:
Looks good to me, thanks!

Issue #4, opened 2018-05-02
Invalid Signature


dsanders11:
Trying to get this library working with a [Yubico Security Key](https://www.yubico.com/product/security-key-by-yubico/) (not the 2 version which supports FIDO2, that seems fairly recent). I keep running into `Invalid signature received.` when testing registration, even in the Flask demo. I'm digging into it now, but haven't had any luck so far. From what I can tell it shouldn't be a problem with usage as the security key is returning an attestation, but the library thinks the signature is wrong.

futureimperfect:
Fixed in #9.

PR #9, opened 2018-05-02
Fix hashing clientDataJson, hashAlgorithm is no longer in the spec

dsanders11:
This fixes invalid signature problems as mentioned in issue #4. It looks like the spec has changed and `clientDataJson` no longer specifies a hash algorithm, causing the existing code to always hash an empty string. I tried to keep the change as small and concise as possible, but someone with more knowledge may want to look at it if `_get_client_data_hash` should simply always use `SHA-256` now as the spec only mentions the client data JSON as being hashed with that algorithm.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -891,9 +891,7 @@ def _get_client_data_hash(client_data, decoded_client_data):
 
     hash_alg = client_data.get('hashAlgorithm')
 
-    if not hash_alg:
-        return ''
-    if hash_alg == 'SHA-256':
+    if not hash_alg or hash_alg == 'SHA-256':
         return hashlib.sha256(decoded_client_data).digest()
     elif hash_alg == 'SHA-512':
         return hashlib.sha512(decoded_client_data).digest()


futureimperfect:
Thanks for the fix! Yeah, so the spec [now specifies](https://github.com/w3c/webauthn/pull/710/files) `SHA-256` as the algorithm to use when computing the hash over `cData`. Previously, the algorithm to use was specified in `cData`. I need to go back and update the registration and assertion steps slightly to match the latest version of the spec.

For now, if you get a second can you update your PR to just always use `SHA-256` for computing the `cData` hash? Something like the following should suffice.

```python
def _get_client_data_hash(decoded_client_data):
    if not isinstance(decoded_client_data, six.string_types):
        return ''

    return hashlib.sha256(decoded_client_data).digest()
```

Thanks again for all the PRs!

dsanders11:
> Thanks again for all the PRs!

No problem, I was glad to find an existing Python library that had already done most of the hard work implementing the spec. :smile:

PR #14, opened 2018-11-03
Make RS256 one of the pubKeyCredParms for reg

dsanders11:
Currently the two `pubKeyCredParams` provided for registration options are 'ES256' and -7.

1. I don't believe string values for `alg`, such as 'ES256' are valid, so it's likely being ignored by browsers
2. -7 is the identifier for 'ES256' anyway, so it's duplicated

Most all examples (including the WebAuthn spec) provide -7 (ES256) and -257 (RS256) as `pubKeyCredParams`.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -101,7 +101,7 @@ class WebAuthnMakeCredentialOptions(object):
             },
             'pubKeyCredParams': [
                 {
-                    'alg': 'ES256',
+                    'alg': -257,
                     'type': 'public-key',
                 },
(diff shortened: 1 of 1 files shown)


dsanders11:
Actually, it looks like the library doesn't support RS256 at the moment, so it probably shouldn't be provided. I think 1 & 2 are still both valid, though, so perhaps this PR should instead remove the duplicate so there's only one value in `pubKeyCredParams`?

dsanders11:
Closing this in favor of #16.

PR #15, opened 2018-11-03
Only include user icon URL in reg if provided

dsanders11:
Newer versions of Chrome (68+) don't seem to like a user's `icon` being set to `null`. I'd been passing `None` to `WebAuthnMakeCredentialOptions` for the icon URL, and this used to work fine (with it showing up in the JSON as `null`) but now it causes a silent failure in Chrome, where Chrome's UI initiates the authenticator process, but never sends the message to the authenticator. I tracked this down to the `icon` being `null`. I'll open an issue with Chromium as the silent failure is definitely a bug.

In the mean-time, I don't think `null` is an allowed value for `icon` anyway in the spec, so only include it in the registration dict if it's provided.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -96,8 +96,7 @@ class WebAuthnMakeCredentialOptions(object):
             'user': {
                 'id': self.user_id,
                 'name': self.username,
-                'displayName': self.display_name,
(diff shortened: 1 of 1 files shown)


futureimperfect:
Looks good to me, thanks @dsanders11!

PR #16, opened 2018-11-03
Add Packed Attestation Format (Large Refactor)

dsanders11:
This includes PR #15.

So, this started as what the title suggests, just adding the `packed` attestation format, as I found that my newer Yubico security key (which supports FIDO2) was now returning that format.

However, as I got into it there were a few thing I needed to refactor to make this fit. Mainly:
* Pull the verify signature logic into it's own function since this is reused in various spots. This has the added benefit that it supports three algorithms mentioned in the spec now: ES256, RS256, and PS256.
* Pull the logic for loading a COSE encoded key into a `cryptography` public key out into it's own function, as it was duplicated in several places (with slightly varying implementations) and drop `_decode_public_key`
* `_verify_attestation_statement` now returns the raw public key data (COSE format) from the `auth_data`. Before it was returning an encoded format specific to the U2F attestation.
* Breaking change: The public key value returned in `WebAuthnCredential` is now the base-64 encoded COSE key format. This was necessary to support algorithms other than ES256.

The last bit means any existing implementations which have stored public keys wouldn't work with this PR. A hack could be made to support them (try the new load, if that fails, try the old `_decode_public_key`). The old method stored the public key in a manner that only worked with ES256, so that had to get broken at some point to support all algorithms.

The packed attestation format is fully implemented minus the ECDAA attestation type. However, I've only tested the go right path for my security key, untested bits are RS256, PS256, and self-attestation.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from __future__ import print_function
 
 import base64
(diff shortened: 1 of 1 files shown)


futureimperfect:
Thanks for pulling this together, @dsanders11! It's been on my todo list for awhile, so I really appreciate your contribution! I'll take a look at this sometime over the next few days, but I just wanted to let you know ahead of time how grateful I am. 💯 🥇

futureimperfect:
Hi @dsanders11,

Out of curiosity, did you test these changes with the included Flask demo? Just wondering because I'm getting the following error when I try to run the demo with a non-FIDO2 U2F token:

```
Registration rejected. Error: 'alg'.
```

I will dig in further, but I wanted to check with you first.

Thanks again!

dsanders11:
@futureimperfect, I didn't test it with the Flask demo, just my own implementation. I also didn't try it with a non-FIDO2 U2F token, as I didn't have one handy. I will have one next week though, so I could try it out then.

PR #18, opened 2018-11-03
Use constant time comparison for verification

dsanders11:
Took this idea from the [Yubico FIDO 2](https://github.com/Yubico/python-fido2) Python library. Use constant time comparisons to prevent timing attacks.

It's probably smartest to put an expiration time on challenges to prevent giving attackers an infinite number of tries anyway, but since the library user in this case controls the challenge and it's storage/expiration that can't be guaranteed, so constant time comparison are a nice extra bit of hardening.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -7,11 +7,14 @@ import os
 import struct
 import sys
 
+from builtins import bytes
+
 import cbor2
 import six
 
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import constant_time
 from cryptography.hazmat.primitives.asymmetric.ec import ECDSA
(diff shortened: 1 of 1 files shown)


futureimperfect:
Thanks! Do you mind adding back the change to `requirements.txt`? Building the Docker image fails without it:

```
Step 7/8 : RUN /app/flask_demo/create_db.py
 ---> Running in 31cff39dcaf3
Traceback (most recent call last):
  File "/app/flask_demo/create_db.py", line 3, in <module>
    from app import app
  File "/app/flask_demo/app.py", line 20, in <module>
    from context import webauthn
  File "/app/flask_demo/context.py", line 6, in <module>
    import webauthn  # NOQA
  File "/app/webauthn/__init__.py", line 2, in <module>
    from .webauthn import WebAuthnAssertionOptions
  File "/app/webauthn/webauthn.py", line 10, in <module>
    from builtins import bytes
ImportError: No module named builtins
```

PR #29, opened 2019-03-07
Hyper fido certs

agesagem:
The CA certs enable u2f usb devices from HyperSecu
Also the 'internal' in acceptable_credential produces errors on login - without it works
Thanks for sharing this great repo!


futureimperfect:
Hi @agesagem,

Thanks for the PR! Do you mind giving a few more details about what happens when `internal` is included in the `transports` list?

agesagem:
Hi,
I just re-tested it and it works with internal.
I can no longer see why I came to the conclusion that internal does not work back then.
So forget this PR.
Sorry for the hassle.

futureimperfect:
Hey no worries @agesagem! Perhaps you were previously on an older version of Chrome or Firefox and it was updated? Either way, if you get a chance to update the PR to revert that change, I'd be happy to merge the added CA certs in.

Thanks!

futureimperfect:
Closing this PR since it was resolved in #31.

Issue #34, opened 2019-04-24
AT_NONE attestation type not handled consistently


pthimon:
When using "none" as the attestation format, the verify function will accept this format all the way through to step 16, but then is missing the format from the elif statements, so it will always raise `RegistrationRejectedException('Unknown attestation type.')`, see: https://github.com/duo-labs/py_webauthn/blob/8e0a7ecdbd2138ff52bb51db25442328b4b7d2bb/webauthn/webauthn.py#L726-L745

futureimperfect:
Fixed by #35.

PR #35, opened 2019-04-24
Handle (ignore) the AT_NONE attestation type during step 16

pthimon:
Fixes issue #34


futureimperfect:
LGTM, thanks @pthimon!

PR #36, opened 2019-05-11
Add support for multiple acceptable credentials

mdedonno1337:
Add the support for multiple credentials in the WebAuthnAssertionOptions() object.
This allow the user to use any key to login (when multiple keys are registered on the server side, of course).

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -150,19 +150,22 @@ class WebAuthnAssertionOptions(object):
         if not self.challenge:
             raise AuthenticationRejectedException('Invalid challenge.')
 
-        # TODO: Handle multiple acceptable credentials.
-        acceptable_credential = {
-            'type': 'public-key',
-            'id': self.webauthn_user.credential_id,
-            'transports': ['usb', 'nfc', 'ble', 'internal']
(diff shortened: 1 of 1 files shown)


mdedonno1337:
wrong email used in the first commit, sorry for the force-push ...

futureimperfect:
Thanks for the PR, @mdedonno1337! I just realized that I responded to #39 before taking a closer look at this PR and I see they're very similar. I like that this one isn't API-breaking, but I see why @woodruffw made the changes to the constructor to accept multiple `webauthn_users`. I'm curious to hear your thoughts on this. Thanks!

woodruffw:
Ah, that's what I get for not checking the open PRs...

mdedonno1337:
I've considered the changes on the constructor with multiples `webauthn_users`. The issue for me is the construction of multiple `webauthn_users` objects, with potentials conflicting information (username, display name, ...) that will not be used anyway. That I've "juste" added the check on the use of the `webauthn_user` object to check if the credentials object is a list or not.

futureimperfect:
As mentioned in #39, I'm going to close this PR. Thanks!

PR #37, opened 2019-05-11
Set the timeout as object parameter

mdedonno1337:
Set the timeout value as parameter in the WebAuthnMakeCredentialOptions and WebAuthnAssertionOptions objects, to be able to change it.
A customizable timeout is interessting to use fallback error quicker on the login page (with jquery for example).


futureimperfect:
Looks good to me! Thanks for the PR, @mdedonno1337!

PR #38, opened 2019-05-11
Add default value for all arguments in the WebAuthnUser class

mdedonno1337:
I was asking myself why all arguments in the WebAuthnUser class are mendatory, but not always used.
I had this code working on production:

```
webauthn_user = webauthn.WebAuthnUser( 
    None, None, None, None,
    key[ 'credential_id' ], None, None, config.RP_ID
)
webauthn_assertion_options = webauthn.WebAuthnAssertionOptions( webauthn_user, challenge )
```

This is not a definitive PR, but a snippet to discuss (of course backwards-compatible ; need some checks in the library)

What do you think ?

Edit:
Same comment with this code:

```
webauthn_user = webauthn.WebAuthnUser( 
    None, session[ 'username' ], None, None,
    user[ 'credential_id' ], user[ 'pub_key' ], user[ 'sign_count' ], config.RP_ID
)

webauthn_assertion_response = webauthn.WebAuthnAssertionResponse( 
    webauthn_user,
    assertion_response,
    challenge,
    config.ORIGIN,
    uv_required = False
)
```

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -175,8 +175,20 @@ class WebAuthnAssertionOptions(object):
 
 
 class WebAuthnUser(object):
-    def __init__(self, user_id, username, display_name, icon_url,
-                 credential_id, public_key, sign_count, rp_id):
+    def __init__(self, *args, **kwargs):
+        if len(args) == 8:
+            user_id, username, display_name, icon_url, \
+                credential_id, public_key, sign_count, rp_id = args
+        else:
+            user_id = kwargs.get("user_id", None)
+            username = kwargs.get("username", None)
+            display_name = kwargs.get("display_name", None)
+            icon_url = kwargs.get("icon_url", None)
+            credential_id = kwargs.get("credential_id", None)
+            public_key = kwargs.get("public_key", None)
+            sign_count = kwargs.get("sign_count", None)
+            rp_id = kwargs.get("rp_id", None)
+        
         self.user_id = user_id
         self.username = username
         self.display_name = display_name
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -186,8 +186,8 @@ class WebAuthnAssertionOptions(object):
 
 
 class WebAuthnUser(object):
-    def __init__(self, user_id=None, username=None, display_name=None, icon_url=None,
-                 credential_id=None, public_key=None, sign_count=None, rp_id=None):
+    def __init__(self, user_id, username, display_name, icon_url,
+                 credential_id, public_key, sign_count, rp_id):
         
         if not credential_id:
             raise WebAuthnUserDataMissing("credential_id missing")
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -1066,13 +1066,16 @@ class WebAuthnAssertionResponse(object):
             #             or not, is Relying Party-specific.
             sc = decoded_a_data[33:37]
             sign_count = struct.unpack('!I', sc)[0]
-            if sign_count:
-                if self.webauthn_user.sign_count == None:
-                    raise WebAuthnUserDataMissing("sign_count missing")
 
-                if sign_count <= self.webauthn_user.sign_count:
-                    raise AuthenticationRejectedException(
-                        'Duplicate authentication detected.')
+            if not sign_count:
+                raise AuthenticationRejectedException('Unable to parse sign_count.')
+
+            if not self.webauthn_user.sign_count:
+                raise WebAuthnUserDataMissing('sign_count missing from WebAuthnUser.')
+
+            if sign_count <= self.webauthn_user.sign_count:
+                raise AuthenticationRejectedException(
+                    'Duplicate authentication detected.')
 
             # Step 18.
             #
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -200,10 +200,10 @@ class WebAuthnUser(object):
             sign_count = kwargs.get("sign_count", None)
             rp_id = kwargs.get("rp_id", None)
         
-        if credential_id == None:
+        if not credential_id:
             raise WebAuthnUserDataMissing("credential_id missing")
 
-        if rp_id == None:
+        if not rp_id:
             raise WebAuthnUserDataMissing("rp_id missing")
 
         self.user_id = user_id
@@ -869,7 +869,7 @@ class WebAuthnAssertionResponse(object):
             # If credential.response.userHandle is present, verify that the user
             # identified by this value is the owner of the public key credential
             # identified by credential.id.
-            if self.webauthn_user.username == None:
+            if not self.webauthn_user.username:
                 raise WebAuthnUserDataMissing("username missing")
             
             user_handle = self.assertion_response.get('userHandle')
@@ -889,7 +889,7 @@ class WebAuthnAssertionResponse(object):
             if not isinstance(self.webauthn_user, WebAuthnUser):
                 raise AuthenticationRejectedException('Invalid user type.')
 
-            if self.webauthn_user.public_key == None:
+            if not self.webauthn_user.public_key:
                 raise WebAuthnUserDataMissing("public_key missing")
 
             credential_public_key = self.webauthn_user.public_key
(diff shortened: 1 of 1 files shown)


futureimperfect:
Good observation, @mdedonno1337! I think this looks mostly good. What do you think about leaving the arguments in the constructor the same, but just raising if any of the required parameters are missing? I don't think there should be any major API breaking changes with this fix, but I could be missing something. Thanks again for your contributions!

mdedonno1337:
Not sur that all values have to be checked while creating the WebAuthnUser object, since not necessary mendatory depending upon the usage of the object. The a16563f commit deal with this part.

webauthn/webauthn.py
```diff
@@ -175,8 +177,26 @@ def json(self):
 
 
 class WebAuthnUser(object):
-    def __init__(self, user_id, username, display_name, icon_url,
-                 credential_id, public_key, sign_count, rp_id):
+    def __init__(self, *args, **kwargs):
```
futureimperfect: Can we just leave the method definition as-is, then raise if a necessary value is missing?

webauthn/webauthn.py
```diff
@@ -175,8 +177,26 @@ def json(self):
...
+            user_id = kwargs.get("user_id", None)
+            username = kwargs.get("username", None)
+            display_name = kwargs.get("display_name", None)
+            icon_url = kwargs.get("icon_url", None)
+            credential_id = kwargs.get("credential_id", None)
+            public_key = kwargs.get("public_key", None)
+            sign_count = kwargs.get("sign_count", None)
+            rp_id = kwargs.get("rp_id", None)
+        
+        if credential_id == None:
```
futureimperfect: Let's change this to `if not credential_id:`.

webauthn/webauthn.py
```diff
@@ -175,8 +177,26 @@ def json(self):
...
+            raise WebAuthnUserDataMissing("credential_id missing")
+
+        if rp_id == None:
```
futureimperfect: Same here, `if not rp_id:`.

webauthn/webauthn.py
```diff
@@ -827,6 +847,9 @@ def verify(self):
             # If credential.response.userHandle is present, verify that the user
             # identified by this value is the owner of the public key credential
             # identified by credential.id.
+            if self.webauthn_user.username == None:
```
futureimperfect: `if not self.webauthn_user.username:`

webauthn/webauthn.py
```diff
@@ -844,6 +867,9 @@ def verify(self):
             if not isinstance(self.webauthn_user, WebAuthnUser):
                 raise AuthenticationRejectedException('Invalid user type.')
 
+            if self.webauthn_user.public_key == None:
```
futureimperfect: `if not self.webauthn_user.public_key:`

webauthn/webauthn.py
```diff
@@ -1017,7 +1043,10 @@ def verify(self):
             #             or not, is Relying Party-specific.
             sc = decoded_a_data[33:37]
             sign_count = struct.unpack('!I', sc)[0]
-            if sign_count or self.webauthn_user.sign_count:
+            if sign_count:
```
futureimperfect: How about we change this to `if not sign_count or not self.webauthn_user.sign_count:`? Then, you can raise your `WebAuthnUserDataMissing` exception and we can assume the `sign_count` is present further down.

webauthn/webauthn.py
mdedonno1337: The problem would be if not all arguments are set, like in :

```
WebAuthnUser(credential_id = "value")
TypeError: __init__() takes exactly 9 arguments (2 given)
```

Would be possible to set all parameters to None, and check them after :

```
def __init__(self, user_id=None, username=None, display_name=None, icon_url=None,
                 credential_id=None, public_key=None, sign_count=None, rp_id=None)
```
But I dont know if it's more readable.

webauthn/webauthn.py
mdedonno1337: With something like 
```
            sc = decoded_a_data[33:37]
            sign_count = struct.unpack('!I', sc)[0]
            if not sign_count or not self.webauthn_user.sign_count:
                raise WebAuthnUserDataMissing("sign_count missing")

            if sign_count <= self.webauthn_user.sign_count:
                raise AuthenticationRejectedException(
                    'Duplicate authentication detected.')
```
We will have a `WebAuthnUserDataMissing` if the key does not provide a `sign_count` correctly, and the error would be a `WebAuthnUserDataMissing` which could let think that the error would come from the `WebAuthnUser` object (server side).

What do you think ?

webauthn/webauthn.py
futureimperfect: I think it's more idiomatic to leave the expected parameters as positional arguments. The `TypeError` signals to the caller that 9 arguments are expected, which they are. It makes sense to make the optional arguments keyword arguments, and set their default values to `None`, and leave the required arguments as positional. The problem with doing this, however, is that we risk making a non-backwards compatible API change. I think at this point it'd be better to leave the method signature as-is, (except for adding new keyword arguments). If we need to add a new positional argument in the future I think at that point it would make sense to rework the positional arguments and make non-essential ones `kwargs`, (maybe `icon_url`?). For now we can just raise if a required positional argument is empty, (`None` or `''`).

webauthn/webauthn.py
futureimperfect: Good point, how about something like the following?

```
            sc = decoded_a_data[33:37]
            sign_count = struct.unpack('!I', sc)[0]

            if not sign_count:
                raise AuthenticationRejectedException("Unable to parse sign_count.")

            if not self.webauthn_user.sign_count:
                raise WebAuthnUserDataMissing("sign_count missing from WebAuthnUser.")

            if sign_count <= self.webauthn_user.sign_count:
                raise AuthenticationRejectedException(
                    'Duplicate authentication detected.')
```

webauthn/webauthn.py
mdedonno1337: With the current signature :
```
def __init__(self, user_id, username, display_name, icon_url,
                 credential_id, public_key, sign_count, rp_id):
```
and not changing it to a more flexible one, for example :

```
def __init__(self, user_id=None, username=None, display_name=None, icon_url=None,
                 credential_id=None, public_key=None, sign_count=None, rp_id=None):
```
we will have to provide anyway the non-mendatory data like this :

```
webauthn_user = WebAuthnUser( 
    None, None, None, None,
    "credential_id", None, None, "rp_id"
)
```
which was the main argument of the PR.

mdedonno1337:
> Thanks for the updates! When you get a chance, do you mind addressing the changes I requested?

Done (I think).
Consider a squash-merge instead of a normal-merge if you consider it more useful.

webauthn/webauthn.py
```diff
@@ -184,8 +186,15 @@ def json(self):
 
 
 class WebAuthnUser(object):
-    def __init__(self, user_id, username, display_name, icon_url,
-                 credential_id, public_key, sign_count, rp_id):
+    def __init__(self, user_id=None, username=None, display_name=None, icon_url=None,
```
futureimperfect: Hi @mdedonno1337,

Can we get rid of the changes to the constructor here?

futureimperfect: Basically, just leave the old `def __init__(...)` method signature.

futureimperfect:
Sorry for all the back-and-forth, @mdedonno1337! If you remove the changes to the method signature for the `WebAuthnUser` constructor we should be good to go. Thanks again!

webauthn/webauthn.py
mdedonno1337: Well, with this modification, the PR is probable useless since the repeated None will be mendatory in the `WebAuthnUser` creation.

mdedonno1337:
Useless PR finally I think (see the comment https://github.com/duo-labs/py_webauthn/pull/38#discussion_r287563733 ).
Let me know when a refactoring process will be mendatory to add variables to the constructor, I will do it with pleasure.

futureimperfect:
Hey @mdedonno1337,

I still think this PR has value, even without the changes to the constructor. It will `raise` if necessary inputs are not supplied to the `WebAuthnUser` class. If you'd rather wait until I revisit making changes to the constructor, that's totally understandable. Thanks again for your contribution!

mdedonno1337:
It's a pleasure to contribute to a usefull project !
A bit difficult for my first real contribution to opensource via PR because of the back-and-forth, but I'm very happy to contribute.

Can you squash-merge, or I have to rebase ? The git-log will be a mess if not squashed ...

futureimperfect:
Thanks so much for your contribution, @mdedonno1337! Your PR is squashed and merged now. :)

PR #40, opened 2019-05-30
WebAuthnMakeCredentialOptions: configurable attestation conveyance

evilscientress:
I modified `WebAuthnMakeCredentialOptions` to allow the configuration of the attestation conveyance that up until now was hard coded to `direct`.

I need to be able to set it to other values to respect users privacy. 

The interface of the new constructor is backwards compatible and values give to the new attestation parameter are verified to only allow values that are allowed by the specification. If a value is give that is not allowed a `ValueError` is raised.


futureimperfect:
This looks good to me, thanks!

PR #43, opened 2019-07-08
Add support for authenticatorSelection->userVerification field

mdedonno1337:
With the Windows update 1903, the behaviour seems different... With old versions of Windows, I was able to use my keys without any PIN. Since the update, the protocol used seems to be FIDO2 instead of FIDO-U2F, breaking some of my codes. 
After investigation, the `WebAuthnMakeCredentialOptions.registration_dict` property did not contain the `authenticatorSelection->userVerification` field (https://www.w3.org/TR/webauthn/#userVerificationRequirement).

This option allow the server to enforce or not the creation of a PIN (actual Windows behaviour) on the client-side, "e.g., in the interest of minimizing disruption to the user interaction flow" as mentionned in the W3C webauthn Recommendation linked.

This commit should not be api breaking, and the default behaviour is the same as without the commit.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -85,9 +85,11 @@ class WebAuthnUserDataMissing(Exception):
 class WebAuthnMakeCredentialOptions(object):
 
     _attestation_forms = {'none', 'indirect', 'direct'}
+    _userVerification = {'required','preferred','discouraged'}
 
     def __init__(self, challenge, rp_name, rp_id, user_id, username,
-                 display_name, icon_url, timeout=60000, attestation='direct'):
+                 display_name, icon_url, timeout=60000, attestation='direct',
+                 userVerification=None):
         self.challenge = challenge
         self.rp_name = rp_name
         self.rp_id = rp_id
@@ -102,6 +104,11 @@ class WebAuthnMakeCredentialOptions(object):
             raise ValueError('Attestation must be a string and one of ' +
                              ', '.join(self._attestation_forms))
         self.attestation = attestation
+
+        if userVerification != None and userVerification not in self._userVerification:
+            raise ValueError('userVerification must be a string and one of ' +
+                             ', '.join(self._userVerification))
+        self.userVerification = userVerification
 
     @property
     def registration_dict(self):
@@ -138,6 +145,11 @@ class WebAuthnMakeCredentialOptions(object):
                 'webauthn.loc': True
             }
         }
+
+        if self.userVerification:
+            registration_dict['authenticatorSelection'] = {
+                'userVerification': self.userVerification
+            }
 
         if self.icon_url:
             registration_dict['user']['icon'] = self.icon_url
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -105,7 +105,7 @@ class WebAuthnMakeCredentialOptions(object):
                              ', '.join(self._attestation_forms))
         self.attestation = attestation
 
-        if user_verification != None:
+        if user_verification is not None:
             user_verification = str(user_verification).lower()
             if user_verification not in self._user_verification:
                 raise ValueError('user_verification must be a string and one of ' +
@@ -148,7 +148,7 @@ class WebAuthnMakeCredentialOptions(object):
             }
         }
 
-        if self.user_verification:
+        if self.user_verification is not None:
             registration_dict['authenticatorSelection'] = {
                 'userVerification': self.user_verification
             }
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -105,9 +105,11 @@ class WebAuthnMakeCredentialOptions(object):
                              ', '.join(self._attestation_forms))
         self.attestation = attestation
 
-        if userVerification != None and userVerification not in self._userVerification:
-            raise ValueError('userVerification must be a string and one of ' +
-                             ', '.join(self._userVerification))
+        if userVerification != None:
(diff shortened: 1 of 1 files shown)


webauthn/webauthn.py
```diff
@@ -85,9 +85,11 @@ class WebAuthnUserDataMissing(Exception):
 class WebAuthnMakeCredentialOptions(object):
 
     _attestation_forms = {'none', 'indirect', 'direct'}
+    _userVerification = {'required','preferred','discouraged'}
```
futureimperfect: Minor nit: can you put spaces after the commas?

`_userVerification = {'required', 'preferred', 'discouraged'}`

webauthn/webauthn.py
```diff
@@ -85,9 +85,11 @@ class WebAuthnUserDataMissing(Exception):
...
 
     def __init__(self, challenge, rp_name, rp_id, user_id, username,
-                 display_name, icon_url, timeout=60000, attestation='direct'):
+                 display_name, icon_url, timeout=60000, attestation='direct',
+                 userVerification=None):
```
futureimperfect: Instead of `None` here, can we default to one of the options? Maybe `preferred`?

webauthn/webauthn.py
```diff
@@ -103,6 +105,13 @@ def __init__(self, challenge, rp_name, rp_id, user_id, username,
                              ', '.join(self._attestation_forms))
         self.attestation = attestation
 
+        if userVerification != None:
```
futureimperfect: If the `userVerification` default is one of `self._userVerification`, then we don't need this check. Or were you thinking that you didn't want the JSON payload to be different? If that's the case, can we at least say `if userVerification is not None` instead?

webauthn/webauthn.py
```diff
@@ -103,6 +105,13 @@ def __init__(self, challenge, rp_name, rp_id, user_id, username,
...
+            userVerification = str(userVerification).lower()
```
futureimperfect: I don't think we need this line if we make the other change I'm suggesting below.

webauthn/webauthn.py
```diff
@@ -103,6 +105,13 @@ def __init__(self, challenge, rp_name, rp_id, user_id, username,
...
+            if userVerification not in self._userVerification:
+                raise ValueError('userVerification must be a string and one of ' +
+                                 ', '.join(self._userVerification))
+        self.userVerification = userVerification
```
futureimperfect: How about we put this line in an `else:` block? That way, it will `raise` if the provided value isn't one of the supported options, else set the value.

webauthn/webauthn.py
futureimperfect: Minor nit: `_user_verification` instead of `_userVerification`.

webauthn/webauthn.py
futureimperfect: Minor nit: `self.user_verification` instead of `self.userVerification`.

webauthn/webauthn.py
mdedonno1337: The use of `.lower()` make the api more flexible since we can use it with upper case parameters.

webauthn/webauthn.py
mdedonno1337: I was simply coding this part like the check for the attestation, in line 102 in this PR.
I think that uniformity in logic is more important than code cleverness.

On the other hand, it make the api more flexible (see comment https://github.com/duo-labs/py_webauthn/pull/43#discussion_r312464220 also).

webauthn/webauthn.py
mdedonno1337: The use of None instead of a "normal" value is to prevent adding the authenticatorSelection value to the registration_dict if not provided.
I think there is a difference between not setting the value (hence letting the operating system decinding the behaviour) and forcing to the standard value. A change in the standard will be pushed directly in usage if we use None ; if not, we will have to patch this function with the new standard value.

The final behaviour is probably the same.

webauthn/webauthn.py
mdedonno1337: Uniformity, see above

webauthn/webauthn.py
futureimperfect: Fair enough. Let's go ahead and leave the default value as `None`, then.

webauthn/webauthn.py
futureimperfect: Sounds good to me. Can you just change it to `if user_verification is not None:`?

webauthn/webauthn.py
mdedonno1337: I dont want to force in any way, you are the owner of the project!
I juste want to contribute in the best way possible to this awsome project.
PS: And by the way, it's my first (real) open-source contribution (not taking into account my own projects, of course).

webauthn/webauthn.py
mdedonno1337: yeap

webauthn/webauthn.py
futureimperfect: 👍Sounds good. You're right that it makes it more consistent, and we already do this for the `attestation` value.

webauthn/webauthn.py
futureimperfect: 👍

futureimperfect:
Thanks for making the changes I requested and explaining your thinking behind some of the others! If you could just change the line that checks if `user_verification` is `None` to this then I'm good to merge:

`if user_verification is not None:`

Thanks again!

Issue #49, opened 2019-08-18
webauthn_assertion_response.verify incorrect padding on public key


haydenth:
Hello! Great library, I'm using it for some projects I'm working on. I'm having some problems getting the `webauthn_assertion_response.verify` method to work ONLY on the cases where I am completing the assertion of a key. The error is below.

```
pipe_web          | Traceback (most recent call last):
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/webauthn/webauthn.py", line 899, in verify
pipe_web          |     _webauthn_b64_decode(credential_public_key))
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/webauthn/webauthn.py", line 1173, in _webauthn_b64_decode
pipe_web          |     return base64.urlsafe_b64decode(encoded)
pipe_web          |   File "/usr/lib/python3.6/base64.py", line 133, in urlsafe_b64decode
pipe_web          |     return b64decode(s)
pipe_web          |   File "/usr/lib/python3.6/base64.py", line 87, in b64decode
pipe_web          |     return binascii.a2b_base64(s)
pipe_web          | binascii.Error: Incorrect padding
pipe_web          | 
pipe_web          | During handling of the above exception, another exception occurred:
pipe_web          | 
pipe_web          | Traceback (most recent call last):
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2463, in __call__
pipe_web          |     return self.wsgi_app(environ, start_response)
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2449, in wsgi_app
pipe_web          |     response = self.handle_exception(e)
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1866, in handle_exception
pipe_web          |     reraise(exc_type, exc_value, tb)
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/_compat.py", line 39, in reraise
pipe_web          |     raise value
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2446, in wsgi_app
pipe_web          |     response = self.full_dispatch_request()
pipe_web          |   File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1951, in full_dispatch_request
pipe_web          |     rv = self.handle_user_exception(e)
[...]

PR #50, opened 2019-08-18
On py3, convert webauthn_credential.public_key to str

haydenth:
RE: issue #49


futureimperfect:
LGTM, thanks @haydenth!

PR #51, opened 2019-08-30
Fix PS256 and RS256 public key validation

hawi74:
Fix validating PS256 and RS256 public keys. This always failed since modulus is required to be at least 2048 bits long.

Changed the tests a bit to make them run for both ES256 and RS256 algs.


futureimperfect:
LGTM, thanks!

PR #52, opened 2019-09-11
Add Solokeys trusted certificates.

Natim:


futureimperfect:
LGTM. Thanks @Natim!

For posterity, I confirmed that the root certificate added here is in fact [the root CA from the solokeys repository](https://github.com/solokeys/solo/blob/master/metadata/Solo-FIDO2-U2F-Authenticator.json).

PR #57, opened 2019-10-31
Fix isinstance(..., int) check on Python 2.7

dsanders11:
Currently the `isinstance` check for `sign_count` will fail on Python2.7 if it is a `long`. This change makes that check pass as expected in both Python 2 and 3.


futureimperfect:
Hi @dsanders11 ,

I tested this change and it looks good to me. @jordan-wright or @nicksteele when you get a chance can you review/merge?

PR #58, opened 2019-11-21
Handle authenticators not supporting the signature counter (Fixes #55).

### LINKED ISSUE #55: Signature counters should be optional.
The WebAuthN spec seems to allow for authenticators which do not support the signature counter (see step 17 in [7.2 Verifying an Authentication Assertion](https://www.w3.org/TR/webauthn/#verifying-asssertion)), i.e. the signature counter should only be checked if it is non-zero. However, the following code:

https://github.com/duo-labs/py_webauthn/blob/a48ab13c06ad2c44cdf226cc6b8483d45b350691/webauthn/webauthn.py#L1081

assumes it is non-zero. I think this line should be preceded by code along the lines of:

```python
           if sign_count == 0 and self.webauthn_user.sign_count == 0:
               return sign_count
```

jonathanverner:
The WebAuthN spec seems to allow for authenticators which do not support the signature counter (see step 17 in [7.2 Verifying an Authentication Assertion](https://www.w3.org/TR/webauthn/#verifying-asssertion)), i.e. the signature counter should only be checked if it is non-zero.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -1077,7 +1077,10 @@ class WebAuthnAssertionResponse(object):
             #             or not, is Relying Party-specific.
             sc = decoded_a_data[33:37]
             sign_count = struct.unpack('!I', sc)[0]
-
+            
+            if sign_count == 0 and self.webauthn_user.sign_count == 0:
+                return sign_count
+            
             if not sign_count:
                 raise AuthenticationRejectedException('Unable to parse sign_count.')
 


skorokithakis:
What is the status of this PR? It seems like a straightforward change.

futureimperfect:
Hi @jonathanverner and @skorokithakis,

I tested this change and it looks good to me. @jordan-wright or @nicksteele when you get a chance can you review/merge?

skorokithakis:
I would possibly change the `return` to `return 0` to be more explicit, but it looks good to me too.

Issue #54, opened 2019-10-17
Wrong salt length when checking PS256 signatures


jonathanverner:
As I understand it the COSE PS256 signature algorithm  is specified in [RFC8230](https://tools.ietf.org/html/rfc8230) (see [Cose Algorithms](https://www.iana.org/assignments/cose/cose.xhtml#algorithms)) and there it states that salt length should be 32 . However, the `_verify_signature` method from `webauthn.py` uses `PSS.MAX_SIZE` :

https://github.com/duo-labs/py_webauthn/blob/a48ab13c06ad2c44cdf226cc6b8483d45b350691/webauthn/webauthn.py#L1365

which leads to signatures failing to verify.

futureimperfect:
Hi @jonathanverner,

It looks like you’re correct. Can you verify that signatures are verifying successfully if you set `salt_length` to `32`? I suspect this was set to `PSS.MAX_LENGTH` because that’s what’s recommended in the Cryptography module’s documentation, but I believe @dsanders11 made the change so perhaps he knows better.

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.padding.PSS

dsanders11:
@futureimperfect, unfortunately that PR was long enough ago that any specific details in my memory are long gone. I was basing it on a combination of sources, so it's very likely something wa just lost in translation, and I didn't have a device to test PS256 with.

jonathanverner:
@futureimperfect Sorry it took me so long, I was swamped with other work. Anyway, yes I checked that the signatures are validating correctly with salt length set to 32 (I've also checked the demo at https://webauthn.io/ and there it works too). I've created a [pull request](https://github.com/duo-labs/py_webauthn/pull/59).

nickmooney:
Fixed by merging #59! Thanks all.

PR #59, opened 2019-11-21
Fix salt length for PS256 signatures (Fixes Issue #54).

jonathanverner:
The COSE PS256 signature algorithm is specified in
[RFC8230](https://tools.ietf.org/html/rfc8230) (see Cose
[Algorithms](https://www.iana.org/assignments/cose/cose.xhtml#algorithms))
where it is stated that salt length should be 32 .


futureimperfect:
Hi @jonathanverner,

Thanks for submitting this PR, and sorry for the delay. I tested this change and it looks good to me, but unfortunately I'm not able to merge it because I'm no longer a member of Duo Labs. Perhaps @jordan-wright or @nicksteele can review and merge it. Thanks!

PR #62, opened 2020-01-13
Add the parameter 'userVerification' to the WebAuthnAssertionOptions …

mdedonno1337:
I have noticed today that my js console has a warning regarding non set userVerification field:
"publicKey.userVerification was not set to any value in Web Authentication navigator.credentials.get() call. This defaults to 'preferred', which is probably not what you want. If in doubt, set to 'discouraged'. See https://chromium.googlesource.com/chromium/src/+/master/content/browser/webauth/uv_preferred.md for details."

This RP should remove this warning.

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -162,13 +162,14 @@ class WebAuthnMakeCredentialOptions(object):
 
 
 class WebAuthnAssertionOptions(object):
-    def __init__(self, webauthn_user, challenge, timeout=60000):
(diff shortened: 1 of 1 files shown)


danielperna84:
I can confirm this removes the error.

nickmooney:
I agree that the default behavior here should be to generate non-resident credentials. I'm going to merge this PR at the moment, but it's worth looking in the future at how py_webauthn supports a resident-key flow (especially in the single factor case).

PR #64, opened 2020-01-19
Fix verification on Python 3 before 3.6

dsanders11:
`json.loads` only supports bytes on Python 3.6+

### DIFF
--- a/webauthn/webauthn.py
+++ b/webauthn/webauthn.py
@@ -599,7 +599,7 @@ class WebAuthnRegistrationResponse(object):
             # parser on JSONtext.
             decoded_cd = _webauthn_b64_decode(json_text)
 
-            if sys.version_info < (3, 0):  # if python2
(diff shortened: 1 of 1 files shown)


nickmooney:
Thanks for the PR, @dsanders11! I'm ready to merge it -- do you know off the top of your head if the calls to `json.loads` on `assertion_client_extensions` and `registration_client_extensions` might also be affected by this issue?

dsanders11:
@nickmooney, quite possibly. Since those values are supplied by the user of the library, though, it's less of an issue as the decode can be done by the developer. The ones in this PR are internal to the library and can't be worked around.

It wouldn't hurt to do a check like in `_webauthn_b64_decode` and decode if needed, but I'm not sure it's necessary.

nickmooney:
Good stuff. Thank you for the PR!

PR #68, opened 2020-02-27
Flask demo client-side js 

gradyy:
`getCredentialCreateOptionsFromServer` and `getCredentialRequestOptionsFromServer` differ only by a hardcoded url. The same applies to `postAssertionToServer` and `postNewAssertionToServer`. These functions can be consolidated into functions that takes the url as a parameter.

This change is to improve readability. The destination urls have been moved to the main body of `didClickRegister` and `didClickLogin` to help the reader more easily identify the server-side function being invoked at each step of the registration or authentication process. In short, the information is all in one place. Having less code to review should also be a plus for the readability factor. 

A future improvement may be to use Flask templating and [url_for](https://flask.palletsprojects.com/en/1.1.x/api/#flask.url_for) to insert the appropriate urls at runtime. 

commits prior to [da4a230](https://github.com/duo-labs/py_webauthn/commit/da4a230f473321eff411452b1b764408f00b7497) are duplicated from #67

### DIFF
--- a/flask_demo/static/js/webauthn.js
+++ b/flask_demo/static/js/webauthn.js
@@ -100,7 +100,7 @@ const transformCredentialRequestOptions = (credentialRequestOptionsFromServer) =
     let {challenge, allowCredentials} = credentialRequestOptionsFromServer;
 
     challenge = Uint8Array.from(
-        atob(challenge), c => c.charCodeAt(0));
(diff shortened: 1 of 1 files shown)


gradyy:
one commit duplicated from #73 
removed unused copy of JQuery.

PR #71, opened 2020-03-13
Fix demo challenge randomness, restructure tests

nickmooney:
In the Flask demo, previously, we were generating challenges by randomly pulling from the URLsafe base64 alphabet.

Ideally we should be generating cryptographically random challenges in bytes, and then encoding them. This change does that. The generate_challenge utility function will generate a *padded* base64url-encoded challenge. This challenge is the one that will be shipped down to the browser, and the padding is kept intact for easy decoding. Before the challenge is stored in the session object, it will be stripped of padding, since the padding value we receive from the browser in the form of CollectedClientData will be base64url-encoded with no padding.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


jordan-wright:
Comments given offline, but this LGTM!

PR #77, opened 2020-05-31
feat: Adds base exception class for the module

Darkheir:
This PR adds a common error class to all the exception of the module.

It makes it easier to catch the potential error from outside the module.


MasterKale:
The new `webauthn.helpers.exceptions` module contains a whole slew of new exceptions that can be used for more nuanced handling of errors raised by the newly released v1.0.0 of the library

Issue #106, opened 2021-11-10
SafetyNet attestation too strict for "now < timestamp_ms"


MasterKale:
A recent error was thrown when trying to register an Android device:

> Payload timestamp 1636585252672 was later than 1636585252000

The culprit was SafetyNet verification failing because **the response appeared to come from 672ms in the future**.

I need to bake in some kind of buffer on the `now < payload.timestamp_ms` check [here](https://github.com/duo-labs/py_webauthn/blob/master/webauthn/registration/formats/android_safetynet.py#L135) to allow for a bit of time drift and network latency between the RP and Google's SafetyNet API.

MasterKale:
This should now be fixed as of [py_webauthn@1.1.0](https://pypi.org/project/webauthn/1.1.0/).

PR #107, opened 2021-11-11
safetynet-time-drift

MasterKale:
Natural time drift between RP and Google SafetyNet API is inevitable, even if it's just a few hundred milliseconds. Unfortunately the library is too strict on timing and will error out on a SafetyNet response that appears to be even a millisecond from the future.

This diff addresses this problem by checking that a SafetyNet response originated sometime between "now +/- 10 seconds".

Addresses #106.

### DIFF
--- a/webauthn/helpers/verify_safetynet_timestamp.py
+++ b/webauthn/helpers/verify_safetynet_timestamp.py
@@ -17,5 +17,5 @@ def verify_safetynet_timestamp(timestamp_ms: int) -> None:
         )
 
     # Make sure the response arrived within the grace period
-    if (timestamp_ms + grace_ms) < now:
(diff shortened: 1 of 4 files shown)


webauthn/helpers/verify_safetynet_timestamp.py
```diff
@@ -0,0 +1,21 @@
+import time
+
+
+def verify_safetynet_timestamp(timestamp_ms: int) -> None:
+    """Handle time drift between an RP and the Google SafetyNet API servers with a window of
+    time within which the response is valid
+    """
+    # Buffer period in ms
+    grace_ms = 10 * 1000
```
jwag956, line 9: For a lower level package/library - I would suggest making constants like this configurable

PR #128, opened 2022-04-03
Add credential protection policy and large blob support

### LINKED ISSUE #127: Support CredentialProtectionPolicy
When working with discoverable credentials, it is favourable to be able to control the credential protection policy (https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension). Currently, py_webauthn does not support this.

jeriox:
We needed support for large blobs and credential protection policy in our project, which py_webauthn didn't provide. We added the appropriate structs and options to genereate_registration_options which worked fine for us. 

closes #127

### DIFF
--- a/webauthn/authentication/generate_authentication_options.py
+++ b/webauthn/authentication/generate_authentication_options.py
@@ -2,11 +2,11 @@ from typing import List, Optional
 
 from webauthn.helpers import generate_challenge
 from webauthn.helpers.structs import (
+    AuthenticationExtensionClientInputs,
+    AuthenticationExtensionsLargeBlobInputs,
     PublicKeyCredentialDescriptor,
     PublicKeyCredentialRequestOptions,
     UserVerificationRequirement,
-    AuthenticationExtensionsLargeBlobInputs,
-    AuthenticationExtensionClientInputs,
 )
 
 
--- a/webauthn/authentication/generate_authentication_options.py
+++ b/webauthn/authentication/generate_authentication_options.py
@@ -5,6 +5,8 @@ from webauthn.helpers.structs import (
     PublicKeyCredentialDescriptor,
     PublicKeyCredentialRequestOptions,
     UserVerificationRequirement,
+    AuthenticationExtensionsLargeBlobInputs,
+    AuthenticationExtensionClientInputs,
 )
 
 
@@ -15,6 +17,7 @@ def generate_authentication_options(
     timeout: int = 60000,
     allow_credentials: Optional[List[PublicKeyCredentialDescriptor]] = None,
     user_verification: UserVerificationRequirement = UserVerificationRequirement.PREFERRED,
+    large_blob_extension: Optional[AuthenticationExtensionsLargeBlobInputs] = None
 ) -> PublicKeyCredentialRequestOptions:
     """Generate options for retrieving a credential via navigator.credentials.get()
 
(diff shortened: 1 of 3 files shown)


webauthn/helpers/structs.py
```diff
@@ -475,6 +474,52 @@ class AttestationObject(WebAuthnBaseModel):
...
+        `USER_VERIFICATION_OPTIONAL`
+        `USER_VERIFICATION_OPTIONAL_WITH_CREDENTIAL_ID_LIST`
+        `USER_VERIFICATION_REQUIRED`
+
+    https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension
+    """
+    USER_VERIFICATION_OPTIONAL = 'userVerificationOptional'
+    USER_VERIFICATION_OPTIONAL_WITH_CREDENTIAL_ID_LIST = 'userVerificationOptionalWithCredentialIDList'
+    USER_VERIFICATION_REQUIRED = 'userVerificationRequired'
+
```
MasterKale, line 522: I understand the value of this extension, but my goal is for this library to support only the functionality defined in the WebAuthn spec itself. Since this extension isn't formally defined in https://www.w3.org/TR/webauthn-2/#sctn-defined-extensions then I'm inclined to request that it not be included in this diff.

webauthn/authentication/generate_authentication_options.py
```diff
@@ -39,10 +43,21 @@ def generate_authentication_options(
...
     )
+
+    ########
+    # Set optional values if specified
+    ########
+
+    if large_blob_extension is not None:
+        options.extensions = AuthenticationExtensionClientInputs(
+            large_blob=large_blob_extension
+        )
```
MasterKale, line 61: Can you please write some tests for this new logic?

webauthn/registration/generate_registration_options.py
```diff
@@ -126,4 +135,10 @@ def generate_registration_options(
             authenticator_selection.require_resident_key = True
         options.authenticator_selection = authenticator_selection
 
+    options.extensions = AuthenticationExtensionClientInputs(
+        large_blob=large_blob_extension,
+        credential_protection_policy=credential_protection_policy,
+        enforce_credential_protection_policy=enforce_credential_protection_policy,
+    )
```
MasterKale, line 142: Can you write some tests for this new logic please to ensure the extensions are set when specified? And I'm curious, what would `extensions` out of this method serialize to JSON as if none of the arguments are provided? An empty `{}`? Would they be omitted? I'd like to see `extensions` not get serialized at all if no extensions are specified.

MasterKale (changes requested):

MasterKale:
I neglected to thank you for submitting this PR: thank you! I've been meaning to sit down and think about extension support with a bit more abstract API that wouldn't require you to know all of the values for a given extension.

For example, why not have an optional `require_large_blob_support: bool` argument to `generate_registration_options()`, since technically the only option in `AuthenticationExtensionsLargeBlobInputs` that's valid for registration is `support`? 

And for authentication you can specify either `read` or `write`, but not both. Therefore I could see adding a `write_large_blob: Optional[bytes] = None` and `read_large_blog: bool = False` to `generate_authentication_options()` so that the developer using the method doesn't need to be aware of the intricacies - we can remove the footguns for them.

What do you think about that approach?

webauthn/helpers/structs.py
MasterKale: There's some conversation going on in an issue in the spec repo that's educating me on how my position in which extensions to support may not be reasonable: https://github.com/w3c/webauthn/pull/1703#discussion_r846202858

I'm going to revisit this PR next week when I get back in the office, your `credProtect` contribution will likely make it through in one way or another.

MasterKale:
Closing for now due to inactivity

Issue #131, opened 2022-04-28
Document base64url_to_bytes


WesleyAC:
As a user of this library, it was quite surprising that it knows which JSON keys are bytes, and uses base64 to serialize/deserialize them — figuring that out required reading the code. It would be great if the readme mentioned that this encoding was in use, since it is non-standard and specific to this library.

MasterKale:
Documentation has been added high-up in the README that should address this.

PR #141, opened 2022-11-07
feat/authenticator-attachment

### LINKED ISSUE #140: Add support for authenticatorAttachment in responses
Modern browsers should return `authenticatorAttachment` to indicate the attachment of the authenticator that was used to register or authenticate. I should make this library aware of this as it is useful to know in passkey scenarios, for example to have an idea of when an RP might want to prompt for platform authenticator registration.

MasterKale:
This PR adds `authenticator_attachment` to `RegistrationCredential` and `AuthenticationCredential`. This is technically the addition of a still-in-working-draft WebAuthn L3 spec feature, but modern browsers are starting to return this value and so I don't see a reason _not_ to implement this.

Fixes #140.

## Screenshots

**Chrome (registration):**

Screen Shot 2022-11-07 at 3 17 13 PM

**Safari (authentication):**

Screen Shot 2022-11-07 at 3 18 47 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,19 @@ Two additional helper methods are also exposed:
 
 Additional data structures are available on `webauthn.helpers.structs`. These [Pydantic-powered](https://pydantic-docs.helpmanual.io/) dataclasses are useful for constructing inputs to the methods above, and for providing type hinting to help ensure consistency in the shape of data being passed around.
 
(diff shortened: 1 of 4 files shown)


MasterKale:
I'm including some README content in here as well to try and address #131.

PR #146, opened 2023-02-11
Add re-export of __init__ imports with __all__

warlo:
Had this issue with mypy in strict mode and figured it could make sense for the webauthn module to expose the methods explicitly and skip the `noqa: F401`.

```
error: Module "webauthn" does not explicitly export attribute "generate_authentication_options"; implicit reexport disabled
...
```

### DIFF
--- a/webauthn/__init__.py
+++ b/webauthn/__init__.py
@@ -9,3 +9,12 @@ from .authentication.verify_authentication_response import (
 from .helpers import base64url_to_bytes, options_to_json
 
 __version__ = "1.7.0"
+
+__all__ = [
+    "generate_registration_options",
+    "verify_registration_response",
+    "generate_authentication_options",
+    "verify_authentication_response",
+    "base64url_to_bytes",
+    "options_to_json",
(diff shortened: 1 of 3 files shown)


grcspace311 (commented):
>#insert import:"revert";type:"unknown" not-auth.
GRCy3

MasterKale:
@warlo Thank you for your submission, this is great! This library is only the second Python library I've ever published to PyPI and so tricks like this one are still new to me. And this all makes mypy happy?

Things have been busy lately or I'd have responded earlier. I'll take a closer look soon and review it all properly.

warlo:
No probs! `__all__` defines what the module exports, and also makes it support `from webauthn import *`. 
Similar case: https://github.com/konradhalas/dacite/issues/133

For mypy's case I did not notice anything else, so should be good afaics 👍

MasterKale:
@warlo Thank you for the contribution! This is going out to PyPI as we speak as `webauthn==1.7.1` 🚀

PR #162, opened 2023-07-17
Upgrade to Pydantic v2

### LINKED ISSUE #159: Update project to support Pydantic v2.0+
Pydantic 2.0 is out with breaking changes. They've helpfully provided migration docs here that I can use to get py_webauthn to track the latest evolution of this dependency:

https://docs.pydantic.dev/2.0/migration/

Based on some of the stuff I'm seeing in there this'll probably form the basis for the eventual release of "`webauthn==2.0.0`".

ljodal:
This upgrades the library to use Pydantic v2. As currently implemented this is very much a breaking change. It also requires Pydantic v2 now and I've not looked into the possibility of supporting both v1 and v2.

The gist of the changes are:

1. Add a URLSafeBase64 type and use it for all bytes fields in `WebAuthnBaseModel` subclasses. This is similar to - and based on - Pydantic v2's Base64 type, but handles the URL safe variant and missing padding. Unlike the previous implementation this requires any input to the field to be valid base64, while the old implementation only required that for certain fields.
2. Update the validator on `WebAuthnBaseModel` to the new pydantic interface
3. Replace usage of deprecated pydantic APIs related to JSON encoding/decoding of models

Fixes #159


tests/test_bytes_subclass_support.py
```diff
@@ -15,10 +19,10 @@ def test_handles_bytes_subclasses(self) -> None:
...
 
         class CustomBytes(bytes):
-            def __new__(cls, data: str):
-                data_bytes = base64url_to_bytes(data)
-                self = bytes.__new__(cls, memoryview(data_bytes).tobytes())
-                return self
+            def __new__(cls, data: Union[str, bytes]):
+                return bytes.__new__(
+                    cls, data.encode("ascii") if isinstance(data, str) else data
+                )
```
MasterKale, line 25: What prompted this rewrite? It doesn't obviously seem driven by changes in Pydantic v2 🤔

tests/test_bytes_subclass_support.py
```diff
@@ -56,35 +64,38 @@ def test_handles_memoryviews(self) -> None:
         Ensure support for libraries that leverage memoryviews
         """
 
-        def base64url_to_memoryview(data: str) -> memoryview:
-            data_bytes = base64url_to_bytes(data)
-            return memoryview(data_bytes)
+        def as_memoryview(data: Union[str, bytes]) -> memoryview:
+            return memoryview(data.encode("ascii") if isinstance(data, str) else data)
```
MasterKale, line 68: Similar to above, what prompted this refactor? I can't see it being obviously driven by changes in Pydantic v2, but maybe I'm missing something.

tests/test_generate_authentication_options.py
```diff
@@ -25,7 +26,7 @@ def test_generates_options_with_custom_values(self) -> None:
         options = generate_authentication_options(
             rp_id="example.com",
             allow_credentials=[
-                PublicKeyCredentialDescriptor(id=b"12345"),
+                PublicKeyCredentialDescriptor(id=urlsafe_b64encode(b"12345")),
```
MasterKale, line 29: `b"12345"` and `urlsafe_b64encode(b"12345")` aren't equivalent, the `id` value needs to be `bytes` because it's the `options_to_json` that encodes `bytes` to base64url-encoded `str`ings...maybe I should come back to tests later once I see the changes to the actual modules...

webauthn/helpers/structs.py
```diff
@@ -1,45 +1,101 @@
...
+    def get_json_format(cls) -> Literal["base64"]:
+        """Get the JSON format for the encoded data.
+
+        Returns:
+            The JSON format for the encoded data.
+        """
+        return "base64"
+
+
+URLSafeBase64 = Annotated[bytes, EncodedBytes(encoder=URLSafeBase64Encoder)]
```
MasterKale, line 64: I feel like there's a lot going on here in this one line, what is this doing exactly?

webauthn/registration/formats/tpm.py
```diff
@@ -234,9 +234,11 @@ def verify_tpm(
...
     try:
         # Ignore mypy because we're casting to a known type
-        ext_subject_alt_name: SubjectAlternativeName = cert_extensions.get_extension_for_oid(
-            ExtensionOID.SUBJECT_ALTERNATIVE_NAME
-        ).value  # type: ignore[assignment]
+        ext_subject_alt_name: SubjectAlternativeName = (
+            cert_extensions.get_extension_for_oid(
+                ExtensionOID.SUBJECT_ALTERNATIVE_NAME
+            ).value
+        )  # type: ignore[assignment]
```
MasterKale, line 241: Out of curiosity, are you making these changes or is Black making them?

MasterKale:
I left some comments, but in general I see changes like this...

Screenshot 2023-07-17 at 12 21 46 PM


...and it seems there's a move in this PR to encoding more bytes to base64url. Is there something about Pydantic v2 that might prevent this library from continuing to use `bytes`, and leave the base64url-encoding to a helper method like `options_to_json`?

webauthn/registration/formats/tpm.py
ljodal: All the formatting changes are Black's doing. I see that the project doesn't have black/mypy/etc. set up on CI, so I guess some changes have snuck in without being properly formatted. I can submit a separate PR with formatting changes and CI setup for running black/mypy etc if you want?

webauthn/helpers/structs.py
ljodal: This is defining a type for Pedantic to use, with encoding/decoding logic "annotated" on. This is the new way to do custom field encoding/decoding in Pydantic v2. It's no longer possible to specify custom json encoding logic for the entire model

ljodal:
@MasterKale all of these changes come back to the first point in the PR description:

> Add a URLSafeBase64 type and use it for all bytes fields in WebAuthnBaseModel subclasses. This is similar to - and based on - Pydantic v2's Base64 type, but handles the URL safe variant and missing padding. Unlike the previous implementation this requires any input to the field to be valid base64, while the old implementation only required that for certain fields.

To elaborate a bit more:

Pedantic v2 has removed support for the model-wide `json_encoders`/`json_loads` options for performance reasons. All of that logic is now handled in the new Rust layer, so supporting custom decoding logic was deemed too complex/costly. [From the migration docs](https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel):

> Due to performance overhead and implementation complexity, we have now removed support for specifying `json_encoders` in the model config. This functionality was originally added for the purpose of achieving custom serialization logic, and we think the new serialization decorators are a better choice in most common scenarios. However, if your usage of json_encoders is not compatible with the new serialization decorators, please create a GitHub issue letting us know. See [Custom data types](https://docs.pydantic.dev/latest/usage/types/custom/) for further details.

The replacement is a per-field serializers, which is what I've created with the `URLSafeBase64` type, which is modelled after how the pydantic provided `Base64Bytes` type does it. Basically that defines a custom field type, which is stored on the model as a bytes field, but it serialized as a URL safe base64 encoded string.

This again leads to all the changes you see related to base64 encoding. The field now expects the input to be a base64 encoded string (or typically a bytes string, as the conversion to a string is done by pydantic) for all fields where this type is specified. This leads to a lot of changes, but honestly it seems a lot more consistent to me than what was there before where the json decoding logic selectively decoded certain fields as base64 based on the field name.

It's possible to make a different field that that only does base64 handling one way, ie when wither serializing or deserializing. For simplicity and consistency I chose to not do that, but if you'd rather I do that I'm happy to make that change.

ljodal:
I'll take another stab at the serialization. It seems it's possible to define a serializer that's only used when serializing to JSON, which seems more like what we want in this case: https://docs.pydantic.dev/latest/usage/serialization/#custom-serializers

MasterKale:
> I'll take another stab at the serialization. It seems it's possible to define a serializer that's only used when serializing to JSON, which seems more like what we want in this case: https://docs.pydantic.dev/latest/usage/serialization/#custom-serializers

Keep in mind as you pursue this that, given that upgrading to Pydantic v2 seems like a pretty big breaking change, I'm inclined to entertain proposals for big API changes like switching to a `URLSafeBase64` instead of `options_to_json`. One thing I want to keep in mind is the migration from `webauthn==1.x` to `webauthn==2.0`, and try to minimize disruption.

MasterKale:
BTW @ljodal I never thanked you for trying to get something working: Thank you! 🙇

ljodal:
No worries :)

I've made a second attempt in https://github.com/duo-labs/py_webauthn/pull/163, which is much cleaner as the base64 encoding/decoding logic is contained to JSON mode in that PR. Closing this as it was much easier to start from scratch than to update this PR.

PR #163, opened 2023-07-17
Upgrade to Pydantic v2

ljodal:
This upgrades the library to use Pydantic v2. As currently implemented this is very much a breaking change. It also requires Pydantic v2 now and I've not looked into the possibility of supporting both v1 and v2.

The gist of the changes are:

1. Add a `Base64URLBytes` type and use it for all bytes fields in WebAuthnBaseModel subclasses. This is similar to - and based on - Pydantic v2's Base64 type, but handles the URL safe variant of base64 and missing padding. Base64 encoding is only done when in JSON mode, when serializing to/validating from Python nothing is done to the value
2. Update the validator on WebAuthnBaseModel to the new `@field_validator` interface
3. Replace usage of deprecated pydantic APIs related to JSON encoding/decoding of models
4. Update some dependencies, including pydantic
5. The `json_loads_base64url_to_bytes` helper has been removed, as it's no longer needed

Note that I've not updated docs/readme etc yet. I'd like some input on this approach before I head down that road


webauthn/helpers/structs.py
```diff
@@ -19,25 +18,27 @@ class WebAuthnBaseModel(BaseModel):
...
-        alias_generator = snake_case_to_camel_case
-        allow_population_by_field_name = True
+    model_config = ConfigDict(
+        alias_generator=snake_case_to_camel_case,
+        populate_by_name=True,
+    )
 
-    @validator("*", pre=True, allow_reuse=True)
-    def _validate_bytes_fields(cls, v, field: ModelField):
+    @field_validator("*", mode="before")
```
ljodal, line 31: It's also possible to handle base64 decoding of `bytes` fields here, but I found no way to use the same `"*"` pattern for `@field_serializer` so I wasn't able to make that work both ways. So because of that I decided to use a separate type instead.

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+    bytes,
+    # Parse input strings as base64url encoded values. This assumes that bytes
+    # values have already been base64 deocded.
+    AfterValidator(validate_base64url),
+    # When serializing to JSON, base64 encode the value. In Python mode we
+    # don't do anything
+    PlainSerializer(bytes_to_base64url, return_type=str, when_used="json"),
+    # Specify the JSON schema of the field, which is a string with the base64
+    # format
+    WithJsonSchema({"type": "string", "format": "base64"}),
```
ljodal, line 39: One benefit of having this be a separate type like this is that you also get correct types in the JSON schema :)

ljodal:
@MasterKale Any chance you could take a look at this?

MasterKale:
Hello @ljodal, thanks for the bump. Rest assured I haven't forgotten about this.

I'm a little hesitant to jump wholeheartedly into v2 because if something goes wrong I'm not sure I'd have time to fix the library. I haven't had as many opportunities to work on the Pydantic v2 upgrade as I'd like...

Help me remember, what would upgrading Pydantic to v2 in here mean for projects consuming py_webauthn and potentially still on Pydantic v1? Is pip/Python smart enough to handle installing v1 and v2? I don't usually have to know this as a user of libraries so I don't have great perspective as a library maintainer on potential fallout from upgrading library dependencies.

ljodal:
There's a quick note about it in the migration guide: https://docs.pydantic.dev/2.0/migration/#continue-using-pydantic-v1-features

MasterKale:
I'm curious what you might think about https://github.com/duo-labs/py_webauthn/pull/164 as an interim step, before forcing all users of py_webauthn to upgrade to Pydantic v2 🤔 

Another thought I had is refactoring out Pydantic use. This library would lose runtime type checks, and I'd have to hand-roll logic to go to/from JSON, but it'd also reduce the dependencies count and as a dependency minimalist that's a Good Thing for me.

jwag956:
Or go back to using attrs which provides for strong typing and self-documentation, without the runtime checks.

martynsmith:
@MasterKale Yeah, it'd be nice to just not have the dependency at all. If you're pressed for time I could probably throw together a version that just removes Pydantic entirely (mostly because I really wanna install v2 in my project but this is currently holding me back) :-)

martynsmith:
Actually, reading through a bit more of the code, pydantic does seem to be a reasonable option (you'd have a lot more code without it). The argument of people being forced to use v1 though is interesting, they'd only be forced to use v1 interacting with this module (which they probably wouldn't even notice given they're just using the classes you've already made) and they'd be free to use v2 throughout the rest of their project?

ljodal:
#164 would cause some additional headaches for us upgrading our project to Pydantic v2, as mentioned in the PR, but at least it would unblock the dependency itself and I could try to solve the new problems.

Could another option be to release this (or some other actual upgrade to v2 variant) as an alpha? That would allow users to upgrade, but with the caveat that it's not "officially" supported and to expect potentially breaking changes.

jwag956:
If pydantic is truly as globally used as it says - then there will be either a huge pushback or quick acceptance. Possibly the easiest thing to do is for a short time support 2 versions of py_webauthn - with different major release numbers - there aren't that many changes happening. That keeps code simple with less possibility of introducing bugs in one or the other. If the 1000's of other packages that use pydantic move quickly - then you can officially stop supporting the V1 version.
Taking the effort to have a single package support both doesn't seem to buy much since from what I have read - you can't actually have both versions installed at the same time anyway. Packaging metadata can make sure that pip installs the correct version based on what the application has requested.

MasterKale:
> Possibly the easiest thing to do is for a short time support 2 versions of py_webauthn - with different major release numbers

"py_webauthn that uses Pydantic v2" would definitely involve a major version bump. `webauthn==1.9.0` would be the last version to support Pydantic v1 (because it already essentially pins to the last v1 release of Pydantic) while `webauthn==2.0.0` would include changes like the ones proposed in this diff. At least that's how I've seen this all going down.

What's still up in the air with me is if there's an attempt at a "1.9.1" version that tries to survive being usable in projects that could be using either v1 or v2 while finalizing the plan to get this library onto v2.

As I type it out I'm not liking the complexity of supporting such a "1.9.1" release. I think this project should have the typical 1.9.0 => 2.0.0 release, with the major version change indicating breaking API changes (in so far as we can't mask most of them with refactors to the various helper methods I emphasize use of in the example.)

🤔

martynsmith:
Yeah, I still support the idea of webauthn supporting both versions of pydantic. I'm definitely in a situation right now where I'd really like to update pydantic to v2 in my own project but I'm restricted because it'll break webauthn.

webauthn/helpers/structs.py
```diff
@@ -19,25 +18,27 @@ class WebAuthnBaseModel(BaseModel):
     - Encodes bytes to Base64URL
     - Converts snake_case properties to camelCase
 
-    `Model.parse_raw()` (from JSON):
+    `Model.model_validate_json()` (from JSON):
```
MasterKale, line 21: Does the reference to `modelInstance.json()` above need to get updated too to mention `model_dump_json()` instead?

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+            return urlsafe_b64decode(value + b"====")
+        except ValueError as e:
+            raise PydanticCustomError(
+                "base64_decode", "Base64 decoding error: '{error}'", {"error": str(e)}
+            ) from e
+
+    return value
+
+
+Base64URLBytes = Annotated[
```
MasterKale, line 29: `Base64URLBytes` seems confusing to me because at first glance "is it string bytes? Or maybe just a byte array?" 

The fact that this gets encoded and decoded between base64url and bytes feels lost in the name. I keep trying to think of a better name for this, though, and don't have any suggestions yet. 🤔

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def validate_base64url(value: bytes, info: FieldValidationInfo) -> bytes:
+
+    if info.mode == "json":
+        try:
+            return urlsafe_b64decode(value + b"====")
+        except ValueError as e:
+            raise PydanticCustomError(
+                "base64_decode", "Base64 decoding error: '{error}'", {"error": str(e)}
```
MasterKale, line 23: I'd like to try and remove as much ambiguity as to whether base64 or base64url is in play. I specifically chose base64url for everything because the spec declares it as a dependency:

https://www.w3.org/TR/webauthn-2/#sctn-dependencies

Thus in places like here I want care to be taken to ensure that we're using base64url through and through.

How about something like this instead:

```suggestion
                "base64url_decode", "Base64URL decoding error: '{error}'", {"error": str(e)}
```

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+            ) from e
+
+    return value
+
+
+Base64URLBytes = Annotated[
+    bytes,
+    # Parse input strings as base64url encoded values. This assumes that bytes
+    # values have already been base64 deocded.
```
MasterKale, line 32:
```suggestion
    # Parse input strings as base64url encoded values. This assumes that bytes
    # values have already been base64url-decoded.
```

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+    AfterValidator(validate_base64url),
+    # When serializing to JSON, base64 encode the value. In Python mode we
+    # don't do anything
```
MasterKale, line 35:
```suggestion
    # When serializing to JSON, base64url-encode the value. In Python mode we
    # don't do anything
```

> In Python mode we don't do anything

I'm not sure what you mean by this, can you please help me understand?

webauthn/helpers/base64url_bytes.py
```diff
@@ -0,0 +1,40 @@
...
+    PlainSerializer(bytes_to_base64url, return_type=str, when_used="json"),
+    # Specify the JSON schema of the field, which is a string with the base64
+    # format
```
MasterKale, line 38:
```suggestion
    # Specify the JSON schema of the field, which is a base64url-encoded string
```

MasterKale:
I opted for PR #166 to add intermediate Pydantic V2 support so this isn't needed anymore.

PR #164, opened 2023-07-31
Support both v1 and v2 of Pydantic

martynsmith:
One of the nice parts about pydantic v2 is that they provide backward compability by importing from `pydantic.v1`. This adds support so that regardless of which version of pydantic the project is using, py_webauthn will work.

### DIFF
--- a/webauthn/helpers/decode_credential_public_key.py
+++ b/webauthn/helpers/decode_credential_public_key.py
@@ -1,7 +1,10 @@
 from typing import Union
 
 from cbor2 import decoder
-from pydantic import BaseModel
+try:
+    from pydantic.v1 import BaseModel
+except ImportError:
+    from pydantic import BaseModel
 
 from .cose import COSECRV, COSEKTY, COSEAlgorithmIdentifier, COSEKey
 from .exceptions import InvalidPublicKeyStructure, UnsupportedPublicKeyType
--- a/webauthn/helpers/decode_credential_public_key.py
+++ b/webauthn/helpers/decode_credential_public_key.py
@@ -4,7 +4,7 @@ from cbor2 import decoder
 try:
     from pydantic.v1 import BaseModel
 except ImportError:
-    from pydantic import BaseModel
+    from pydantic import BaseModel  # type: ignore
 
 from .cose import COSECRV, COSEKTY, COSEAlgorithmIdentifier, COSEKey
 from .exceptions import InvalidPublicKeyStructure, UnsupportedPublicKeyType
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-plugins = pydantic.mypy
+plugins = pydantic.v1.mypy
 
 python_version = 3.8
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+annotated-types==0.5.0
 asn1crypto==1.4.0
 black==21.9b0
 cbor2==5.4.2.post1
@@ -11,11 +12,12 @@ pathspec==0.9.0
 platformdirs==2.4.0
 pycodestyle==2.8.0
 pycparser==2.20
-pydantic==1.10.11
+pydantic==2.1.1
+pydantic_core==2.4.0
 pyflakes==2.4.0
 pyOpenSSL==23.2.0
 regex==2021.10.8
 six==1.16.0
 toml==0.10.2
 tomli==1.2.1
-typing-extensions==4.2.0
+typing_extensions==4.7.1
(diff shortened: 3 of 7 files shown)


martynsmith:
I do see there's a branch to upgrade to Pydantic V2, this feels like a nice stepping stone for people using the library where both will work.

webauthn/helpers/decode_credential_public_key.py
```diff
@@ -4,7 +4,7 @@
 try:
     from pydantic.v1 import BaseModel
 except ImportError:
-    from pydantic import BaseModel
+    from pydantic import BaseModel  # type: ignore
```
MasterKale, line 7: Are the `#type: ignore`'s on the v2 imports because mypy, referencing v1 types, would complain about these?

martynsmith: The imports in the `except` block are actually "v1" imports.

If we can't import from `pydantic.v1` (which only exists in v2), then we instead try to import from `pydantic` (assuming that v1 is installed).

Because in the repo itself we have v2 installed, I've put the ignores on the "v1" imports because when v2 is installed that's actually importing the new v2 API which conflicts with the old v1 version.

MasterKale: Oh right `pydantic.v1` is the new import 🤦‍♂️ 

So would this mean users still using Pydantic v1 in their project wouldn't get mypy-assisted type annotations on the methods in this library if they're still on Pydantic v1? 🤔

martynsmith: Hmm, yeah, I think that's probably right (I have no idea how you'd go about supporting that either) :grimacing:

martynsmith: Actually, depending on how much of the internal pydantic stuff is being exposed, I suspect it probably won't be _that_ much of a degraded experience for those folks?

ljodal: > So would this mean users still using Pydantic v1 in their project wouldn't get mypy-assisted type annotations on the methods in this library if they're still on Pydantic v1? 🤔

From what I've seen mypy will also complain about a lot of stuff if you have v2 installed and use the v1 imports, at least if the mypy plugin for pydantic is enabled.

ljodal:
This seems like a nice stepping stone for unblocking downstream projects from upgrading to Pydantic v2, but in my experience using the v1 imports is a bit clumsy in a project where you mainly use v2 features. For example when you have the `pydantic.mypy` plugin installed it will complain about a bunch of things related to the v1 models. Not entirely sure why.

And if this is the approach we go for it should probably be well documented, as any subclasses of `WebAuthnBaseModel` will also be limited to the v1 api and users would have to be aware of that

ljodal:
Here's a real-life example from our codebase where this approach would cause very confusing behavior. The first model would be a pydantic v1 model while the second one a v2 model:

Screenshot 2023-08-02 at 10 38 44

We also have a decorator we use to automatically generate OpenAPI schemas for our APIs, which relies heavily on Pydantic and I was able to simplify the logic a from the improvements in Pydantic v2. However, that's going to crash if it encounters a `v1.BaseModel` because the `TypeAdapter` class in Pydantic v2 doesn't support that 🤔

ljodal:
I've attempted an alternative approach in #166, turns out Pydantic v2 has native support for serializing bytes to base64 encoded string, and it appears to be using the URL safe variant, so that might be a viable path

martynsmith:
@MasterKale - Are you just going to do the split-version thing (i.e. not use this MR) or still undecided?

I'm just sort of hanging out to upgrade to pydantic2 in my own project and trying to figure out how I'm going to proceed (looking for some idea of a timeline on this if you have any idea what that might be)?

MasterKale:
PR #166 caught my attention and I thought that was going to end up being the way to support v1 and v2, but after coming back to this based on your nudge I think this will probably be the intermediate step. Then I can spend some time working on updating to v2 (though I'm hoping https://github.com/pydantic/pydantic/issues/7000 gets some support because it'll make the migration a lot smoother.)

I'll take a look again in the morning, it's till Sunday evening for me. The fact that tests are passing in this branch gives me hope that I can quickly merge and release as v1.10.0 tomorrow, or Tuesday at the latest to give me some wiggle room.

martynsmith:
Awesome, thanks heaps for that 👍

martynsmith:
Hi @MasterKale , any luck having a look at making a v1.10.0?

MasterKale:
@martynsmith Thank you for creating this PR. I've opted for PR #166 to add intermediate Pydantic V2 support so I'm going to close this one out.

PR #166, opened 2023-08-02
Support both Pydantic v1 and v2

ljodal:
This is another attempt at supporting both Pydantic v1 and v2, but without using the v1 module. Instead we configure `WebAuthnBaseModel` slightly differently based on the pydantic version.

This contains a slight behavior change in that trailing `=` in base64 encoded values are not removed, because pydantic v2 doesn't support that. If that is a problem (I'm not familiar with the spec) I can try to see if I can remove them somehow by doing some post-processing of the serialized values provided by pydantic

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
         'asn1crypto>=1.4.0',
         'cbor2>=5.4.2.post1',
         'cryptography>=41.0.1',
-        'pydantic>=1.10.11,<2.0a0',
+        'pydantic>=1.10.11',
         'pyOpenSSL>=23.2.0',
     ]
 )
--- a/tests/test_verify_registration_response.py
+++ b/tests/test_verify_registration_response.py
@@ -60,7 +60,7 @@ class TestVerifyRegistrationResponse(TestCase):
         assert verification.credential_type == PublicKeyCredentialType.PUBLIC_KEY
         assert verification.sign_count == 23
         assert verification.credential_backed_up == False
-        assert verification.credential_device_type == "single_device"
+        assert verification.credential_device_type == 'single_device'
 
     def test_raises_exception_on_unsupported_attestation_type(self) -> None:
         cred_json = {
--- a/tests/test_verify_registration_response.py
+++ b/tests/test_verify_registration_response.py
@@ -2,6 +2,7 @@ import json
 from unittest import TestCase
 
 import cbor2
+from pydantic import ValidationError
 from webauthn.helpers import base64url_to_bytes, bytes_to_base64url
 from webauthn.helpers.exceptions import InvalidRegistrationResponse
 from webauthn.helpers.known_root_certs import globalsign_r2
@@ -59,7 +60,7 @@ class TestVerifyRegistrationResponse(TestCase):
         assert verification.credential_type == PublicKeyCredentialType.PUBLIC_KEY
         assert verification.sign_count == 23
         assert verification.credential_backed_up == False
-        assert verification.credential_device_type == 'single_device'
+        assert verification.credential_device_type == "single_device"
 
     def test_raises_exception_on_unsupported_attestation_type(self) -> None:
         cred_json = {
@@ -87,9 +88,7 @@ class TestVerifyRegistrationResponse(TestCase):
         rp_id = "localhost"
         expected_origin = "http://localhost:5000"
 
-        with self.assertRaisesRegex(
-            Exception, "value is not a valid enumeration member"
-        ):
+        with self.assertRaises(ValidationError):
             verify_registration_response(
                 credential=credential,
                 expected_challenge=challenge,
--- a/webauthn/authentication/verify_authentication_response.py
+++ b/webauthn/authentication/verify_authentication_response.py
@@ -4,7 +4,7 @@ from typing import List, Union
 from cryptography.exceptions import InvalidSignature
 
 from webauthn.helpers import (
-    bytes_to_base64url,
+    base64url_to_bytes,
     decode_credential_public_key,
     decoded_public_key_to_cryptography,
     parse_authenticator_data,
@@ -69,7 +69,7 @@ def verify_authentication_response(
     """
 
     # FIDO-specific check
-    if bytes_to_base64url(credential.raw_id) != credential.id:
+    if credential.raw_id != base64url_to_bytes(credential.id):
         raise InvalidAuthenticationResponse("id and raw_id were not equivalent")
 
     # FIDO-specific check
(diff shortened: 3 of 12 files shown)


tests/test_verify_registration_response.py
```diff
@@ -87,9 +88,7 @@ def test_raises_exception_on_unsupported_attestation_type(self) -> None:
         rp_id = "localhost"
         expected_origin = "http://localhost:5000"
 
-        with self.assertRaisesRegex(
-            Exception, "value is not a valid enumeration member"
-        ):
+        with self.assertRaises(ValidationError):
```
ljodal, line 91: The message is different in v1 and v2 so I just changed it to assert a validation error is raised

webauthn/authentication/verify_authentication_response.py
```diff
@@ -69,7 +69,7 @@ def verify_authentication_response(
     """
 
     # FIDO-specific check
-    if bytes_to_base64url(credential.raw_id) != credential.id:
+    if credential.raw_id != base64url_to_bytes(credential.id):
```
ljodal, line 72: I flipped this around so any trailing `=` would affect the check, I assume that's a safer comparison anyway?

MasterKale:
You've put in a tremendous amount of work trying to make this all work, thank you again. 🙇 

> This contains a slight behavior change in that trailing = in base64 encoded values are not removed, because pydantic v2 doesn't support that.

...so I feel bad asking this, but can you to try to solve the problem of the padding being included in base64url output? Padding is optional in base64url, and I think it's helped avoid the "base64 vs base64url" ambiguity by omitting the `=` in JSON output.

I tried figuring out how to do it, perhaps each of the `BaseModel` instances get a [`@field_serializer`](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer)? I guess there could be the same `if PYDANTIC_V2` check in models that have these defined so that we don't try to define them when only Pydantic v1 is available?

MasterKale:
I put in a feature request for Pydantic v2 to propose the addition of a "`base64url`" option for `ConfigDict.ser_json_bytes`. I wonder what the response will be...

https://github.com/pydantic/pydantic/issues/7000

ljodal:
> You've put in a tremendous amount of work trying to make this all work, thank you again. 🙇 
> 
> 
> 
> > This contains a slight behavior change in that trailing = in base64 encoded values are not removed, because pydantic v2 doesn't support that.
> 
> 
> 
> ...so I feel bad asking this, but can you to try to solve the problem of the padding being included in base64url output? Padding is optional in base64url, and I think it's helped avoid the "base64 vs base64url" ambiguity by omitting the `=` in JSON output.
> 
> 
> 
> I tried figuring out how to do it with the current version of Pydantic. Perhaps affected instances of `BaseModel` get [`@field_serializer`](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer) definitions for each of their `bytes` fields? I guess there could be the same `if PYDANTIC_V2` check in models that have these defined so that we don't try to define them when only Pydantic v1 is available? But boy, that's a lot more code to maintain...

I can see what's possibly. One option is of course to explicitly define these serializers for every bytes field. That's possible and no big problem, but it'd be a bit of code to maintain as you say. And for any downstream subclasses the developers would have to do the same.


Another alternative is to dynamically add the field serializers. From what I can tell `@field_serialzier` doesn't support specifying `"*"` to get a callback for every field the same way it does for `@field_validator`. So that means at some point, either through a pydantic hook or a Python hook that runs _before_ pydantic initializes the class, we'd need to dynamically inject the serializers for fields declared as bytes. It might be possible to use `[__init_subclass__](https://docs.python.org/3/reference/datamodel.html#customizing-class-creation)`, but I'm not entirely sure.

I'll take a look next week, I'm currently up in the mountains on a little cabin trip without my laptop ⛰️

ljodal:
> I put in a feature request for Pydantic v2 to propose the addition of a "`base64url`" option for `ConfigDict.ser_json_bytes`. I wonder what the response will be...
> 
> 
> 
> https://github.com/pydantic/pydantic/issues/7000

Nice! An alternative is to ask for support for something like this:

```python
class MyBaseModel(BaseModel):
    @field_serializer("*")
    def bytes_to_base64url(self, value: Any, info: SerializerInfo):
        # Check if field is bytes and if so serialize to base64url
```

ljodal:
I got it working with removing trailing `=` @MasterKale. Turns out it wasn't too hard after all. It's done using a model serializer that loops through the fields and strips trailing `=` from any bytes fields that have been serialized as strings:

https://github.com/duo-labs/py_webauthn/blob/ecf34463493da4941c7715d73a9247f9f93dcf09/webauthn/helpers/structs.py#L84-L99

I'm fairly certain CI should be green now, the previous failure was due to a mypy version not supported by the pydantic plugin. Locally I have green mypy and tests on both Pydantic v1 and v2 as this branch currently stands

webauthn/authentication/verify_authentication_response.py
MasterKale: Can you please revert the this change, and the similar one when verifying registration responses? It'll remove the two files completely from this diff, and the rationale for the change got fixed so I don't quite see the need for this to get refactored anymore.

ljodal: Done ✅ What I changed it to seems like a safer choice regardless, but that can be taken in a different PR. If trailing `=` is optional - I'm not familiar with the webauthn spec - the client might send `credential.id` with trailing `=` causing a failed check here. The way I rewrite it trailing `=` or not would not make a difference.

Do you also want me to revert the changes in `webauthn/helpers/bytes_to_base64url.py`? I rewrote the stripping from memory, but not the same way it turns out.

setup.py
```diff
@@ -51,7 +51,7 @@ def find_version(*file_paths):
         'asn1crypto>=1.4.0',
         'cbor2>=5.4.2.post1',
         'cryptography>=41.0.1',
-        'pydantic>=1.10.11,<2.0a0',
```
MasterKale, line 54: Does `pydantic-core` needs to be defined here? I see it got added in **requirements.txt** but not here which is why I ask.

ljodal: It does not. It's a dependency of pydantic, not something that's used directly here. I've updated requirements.txt using `pip freeze > requirements.txt` which is how I'm used to doing it. Happy to remove/add anything you want.

webauthn/authentication/verify_authentication_response.py
MasterKale: I'm fine with the use of `.rstrip("=")` in that method so I'm fine with it getting updated in this PR.

MasterKale:
It was a tough decision, but I'm choosing this PR over #164 because this PR includes the use of V2 when it's present. That should better prepare this library for the eventual refactor to just Pydantic V2 support whenever that becomes viable.

PR #170, opened 2023-08-31
get rid of pydantic v2>v3 deprication warning; add optional indentation

gravmatt:


MasterKale:
Thanks for the contribution @gravmatt, unfortunately this library is aiming to support both v1 and v2 for now so the deprecation warnings can't be addressed like this just yet. If you're interested in updating the PR to work with v1 as well I'll still accept it.

gravmatt:
Hey @MasterKale I understand. webauthn 1.10.1 requires pydantic>=1.10.11. I installed pydantic 1.10.11 and in the `helpers > options_to_json.py` there is the function `options_to_json` which uses `model_dump_json`. This function was introduced in pydantic v2 and is not available in v1. 

https://github.com/duo-labs/py_webauthn/pull/170/commits/2b87a8ce1df06571c448009875f593f72291a9b7#diff-b58c7253734137d5c7558d6204ca581688ed2fed8559c83b28bb7741a8c98172

MasterKale:
@gravmatt Thank you for creating this PR. I've decided to go with #172 because it reduces the library user's exposure to more of the v1/v2 divide. 🙇

PR #172, opened 2023-09-01
avoid reliance on pydantic symbols

### LINKED ISSUE #171: Forced dependency on pydantic for library users
Users of this library need to explicitly rely on symbols that are implemented in pydantic because the public interface expects pydantic instances that have to be built using `{RegistrationCredential, AuthenticationCredential}.parse_raw` (a _v1-ism_) or `.model_validate_json`  (equivalent in v2).

In this way, `py_webauthn` dependency on `pydantic` is "contagious". It's even more complicated because the coupling is to a specific version of `pydantic`. Additionally to the `.parse_raw`/`.model_validate_json` there's also the extra need to depend on `pydantic.ValidationError`.

jpaniagualaconich:
Fixes #171 and #170


webauthn/authentication/verify_authentication_response.py
```diff
@@ -42,7 +43,7 @@ class VerifiedAuthentication(WebAuthnBaseModel):
 
 def verify_authentication_response(
     *,
-    credential: AuthenticationCredential,
+    credential: Union[str, bytes, bytearray, AuthenticationCredential],
```
MasterKale, line 46: I'm curious why you included `bytes` and `bytearray` here. I'm fine with this property gaining the ability to accept raw JSON as `str` which we'll then parse to `AuthenticationCredential`, but I don't immediately see why we'd want to include these other two data types (even if Pydantic has no problem parsing them.)

I have the same question for `verify_registration_response()` too but I'll won't repost this down there.

webauthn/helpers/parse_authentication_credential.py
```diff
@@ -0,0 +1,21 @@
+from typing import Union
+from pydantic import ValidationError
+
+from .exceptions import InvalidAuthenticationResponse
+from .structs import PYDANTIC_V2, AuthenticationCredential
+
+
+def parse_authentication_credential(json_val: Union[str, bytes, bytearray]) -> AuthenticationCredential:
```
MasterKale, line 8: I think this method would better communicate its functionality if it was `parse_authentication_credential_json()` instead. Without the `_json` it almost reads like you'd pass an instance of `AuthenticationCredential` into it, but of course this is just for parsing JSON into a `AuthenticationCredential`.

webauthn/helpers/parse_registration_credential.py
```diff
@@ -0,0 +1,21 @@
+from typing import Union
+from pydantic import ValidationError
+
+from .exceptions import InvalidRegistrationResponse
+from .structs import PYDANTIC_V2, RegistrationCredential
+
+
+def parse_registration_credential(json_val: Union[str, bytes, bytearray]) -> RegistrationCredential:
```
MasterKale, line 8: Similar suggestion as above, I think this method should be called `parse_registration_credential_json()`.

webauthn/authentication/verify_authentication_response.py
jpaniagualaconich: I just didn't want to assume stricter typing was OK but I also don't see uses besides `str`. Fixed.

webauthn/helpers/parse_authentication_credential.py
jpaniagualaconich: I agree. Fixed.

webauthn/helpers/parse_registration_credential.py
jpaniagualaconich: Fixed as well.

webauthn/registration/verify_registration_response.py
```diff
@@ -91,6 +93,8 @@ def verify_registration_response(
     Raises:
         `helpers.exceptions.InvalidRegistrationResponse` if the response cannot be verified
     """
+    if isinstance(credential, (str, bytes, bytearray)):
+        credential = parse_registration_credential_json(credential)
```
MasterKale, line 97: I think we should omit `bytes` and `bytearray` from here as well (and the same above) to align with the changes to the `credential` argument type

webauthn/registration/verify_registration_response.py
```diff
@@ -138,7 +142,7 @@ def verify_registration_response(
                 f'Unexpected token_binding status of "{status}", expected one of "{",".join(expected_token_binding_statuses)}"'
             )
 
-    attestation_object = parse_attestation_object(response.attestation_object)
+    attestation_object = parse_attestation_object(response.attestation_object)  # would be nice to enclose this one in try - except ValidationError and raise InvalidRegistrationResponse
```
MasterKale, line 145: Can you create a new issue about this and make this comment a `# TODO: Issue #NNN` to point to the issue? It's a worthwhile improvement suggestion, but maybe not for this PR.

MasterKale:
mypy is what's failing CI:

```
Run python -m mypy webauthn
webauthn/helpers/parse_registration_credential_json.py:9: error: "Type[RegistrationCredential]" has no attribute "model_validate_json"  [attr-defined]
webauthn/helpers/parse_authentication_credential_json.py:9: error: "Type[AuthenticationCredential]" has no attribute "model_validate_json"  [attr-defined]
webauthn/helpers/options_to_json.py:20: error: Item "PublicKeyCredentialCreationOptions" of "Union[PublicKeyCredentialCreationOptions, PublicKeyCredentialRequestOptions]" has no attribute "model_dump_json"  [union-attr]
webauthn/helpers/options_to_json.py:20: error: Item "PublicKeyCredentialRequestOptions" of "Union[PublicKeyCredentialCreationOptions, PublicKeyCredentialRequestOptions]" has no attribute "model_dump_json"  [union-attr]
webauthn/registration/formats/android_safetynet.py:91: error: "Type[SafetyNetJWSHeader]" has no attribute "model_validate_json"  [attr-defined]
webauthn/registration/formats/android_safetynet.py:92: error: "Type[SafetyNetJWSPayload]" has no attribute "model_validate_json"  [attr-defined]
webauthn/registration/verify_registration_response.py:206: error: "AttestationStatement" has no attribute "model_fields_set"  [attr-defined]
Found 7 errors in 5 files (checked 48 source files)
Error: Process completed with exit code 1.
```

https://github.com/duo-labs/py_webauthn/actions/runs/6052677468/job/16427182539?pr=172

You can run this locally to see what errors it's raising. When mypy reports nothing locally it should all pass in CI too:

```
python -m mypy webauthn
```

webauthn/registration/verify_registration_response.py
jpaniagualaconich: Created

webauthn/registration/verify_registration_response.py
jpaniagualaconich: of course. Fixed.

webauthn/authentication/verify_authentication_response.py
```diff
@@ -67,6 +68,8 @@ def verify_authentication_response(
     Raises:
         `helpers.exceptions.InvalidAuthenticationResponse` if the response cannot be verified
     """
+    if isinstance(credential, (str, bytes, bytearray)):
```
MasterKale, line 71: This will need a similar treatment as below 😊

```suggestion
   if isinstance(credential, str):
```

MasterKale:
@jpaniagualaconich Are you able to see the CI pipeline errors if you click Details above?

webauthn/authentication/verify_authentication_response.py
jpaniagualaconich: and that one too.

jpaniagualaconich:
@MasterKale mypy goes belly up when I install pydantic v1
```
py_webauthn/.venv/lib/python3.11/site-packages/pydantic/env_settings.py:22: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 1.4.1
```

I think latest commit will finally make mypy happy but I can't run it myself.

webauthn/helpers/parse_authentication_credential_json.py
```diff
@@ -0,0 +1,21 @@
...
+        parsing_method: Callable = AuthenticationCredential.model_validate_json  # type: ignore[attr-defined]
+    else:  # assuming V1
+        parsing_method = AuthenticationCredential.parse_raw
+
+    try:
+        authentication_credential = parsing_method(json_val)
+    except ValidationError as exc:
+        raise InvalidAuthenticationResponse(
+            "Unable to parse an authentication credential from JSON data"
+        ) from exc
```
MasterKale, line 19: I'm not familiar with this `from exc` syntax like this in an `except` - does it surface anything from the original exception so users have a chance to understand what specifically was bad?

MasterKale:
This is looking all but ready to go @jpaniagualaconich, thank you again for taking the time to prepare this PR. Just had one final question before I accept.

webauthn/helpers/parse_authentication_credential_json.py
jpaniagualaconich: Exactly. In this way you're explicitly declaring that exception `exc` (an instance of `ValidationError`) was the cause of your new `InvalidAuthenticationResponse` exception.

If you try/except `InvalidAuthenticationResponse` when calling `parse_authentication_credential_json` there will be a `__cause__` attribute that you can inspect if you need to know more about what caused it.

It's called explicit exception chaining. All the info [here](https://peps.python.org/pep-0344/#explicit-exception-chaining).

MasterKale:
Hey @jpaniagualaconich, I know it's been a few days but I do intend to merge this after I get back from TPAC and my schedule clears up a bit to take your changes out for a spin. Thanks for your patience!

MasterKale:
Alright, this looks good to me. Thank you for your contribution @jpaniagualaconich 🎉

PR #174, opened 2023-09-09
Fix handling of optional fields such as user_handle.

### LINKED ISSUE #168: Handling of optional b64 encoded data incorrect
Using pydantic 2.0 - the code in structs::_pydantic_v2_validate_bytes_fields()
has the line:

```
            if field.annotation != bytes:
                return v
```

However for user_handle - the field annotation is: typing.Optional[bytes]
so it simply returns without decoding....

jwag956:
Instead of relying on input types to drive conversion - set up a field_validator that enumerates all spec-defined base64 fields and convert them. Remove the base64url decoding from the generic field_validator.

change (and delete) some tests around use of memoryviews and other bytes equivalents - the idea is that the RP needs the flexibility to store its values in any bytes-equivalent manner - but the responses from the authenticator needs to follow the spec.

closes #168


tests/test_bytes_subclass_support.py
```diff
@@ -63,19 +64,19 @@ def base64url_to_memoryview(data: str) -> memoryview:
...
+                    ,
+                    client_data_json=
                         "eyJ0eXBlIjoid2ViYXV0aG4uZ2V0IiwiY2hhbGxlbmdlIjoiZVo0ZWVBM080ank1Rkl6cURhU0o2SkROR3UwYkJjNXpJMURqUV9rTHNvMVdOcWtHNms1bUNZZjFkdFFoVlVpQldaV2xaa3pSNU1GZWVXQ3BKUlVOWHciLCJvcmlnaW4iOiJodHRwOi8vbG9jYWxob3N0OjUwMDAiLCJjcm9zc09yaWdpbiI6ZmFsc2V9"
-                    ),
-                    signature=base64url_to_memoryview(
+                    ,
+                    signature=
                         "RRWV8mYDRvK7YdQgdtZD4pJ2dh1D_IWZ_D6jsZo6FHJBoenbj0CVT5nA20vUzlRhN4R6dOEUHmUwP1F8eRBhBg"
-                    ),
+                    ,
```
MasterKale, line 79: I don't understand why this test was changed. `base64url_to_memoryview()` emulates an RP that chooses to use a different data structure than plain `bytes`, and then decides to map their data to the shape of `AuthenticationCredential`...

Maybe what's missing here is that `AuthenticationCredential` is an instance of [the spec's `PublicKeyCredential` structure](https://www.w3.org/TR/webauthn-2/#iface-pkcredential), which says `rawId` is an `ArrayBuffer` - that's captured in this library as `raw_id: bytes`. If you want to make it possible for `AuthenticationCredential.raw_id` to be a `str` then I don't think that makes sense, since the library will ensure that `raw_id` bytes can be base64url-encoded to be the same value as `id`. If an RP wants to use something else to store `raw_id` bytes in their database prior to the bytes ever making it to `webauthn.verify_*_response()`, then that should be up to them. They just need to get it back to bytes before this library's methods are called.

jwag956:
Thanks for the comments - I will try tests again once 1.11 is out. I changed the test since my understanding is that these tests are faking up an AuthenticationAssertionResponse - which is a client to RP data structure which the spec dictates the data types for. An RP never creates an AuthenticationAssertionResponse. An RP may, in the case of say user_handle - strore ITS version as a memory view or bytes - and wants to be able to compare that to a properly decoded AuthenticationAssertionResponse - which the current code doesn't properly decode.

The gist of this PR is to say that how client data structures are decoded should be dictated by the spec, not by python/pydantic data structures)

jwag956:
This was an overly-aggressive change - working on something more focused....

PR #183, opened 2023-10-27
Add support for subdomain matches in origins

### LINKED ISSUE #182: Add support for "structural parsing" of expected origins
The webauthn spec on "[Validating the origin of a credential](https://w3c.github.io/webauthn/#sctn-validating-origin)" states that the RP **MUST** validate the origin of a credential, and then lists various validation scenarios. The first two are covered - an exact match on either a single str or a list of str.

The third is not:

> A web application served at a large set of domains that changes often might parse [origin](https://w3c.github.io/webauthn/#dom-collectedclientdata-origin) structurally and require that the URL scheme is https and that the authority equals or is any subdomain of the [RP ID](https://w3c.github.io/webauthn/#rp-id) - for example, example.org or any subdomain of example.org).

This covers the case where a single domain supports multiple subdomains, and where you may want to add additional subdomains which you wish to be covered by the same Passkey without updating your RP settings. (Question - is this a real use case beyond local development?)

Lowest impact change (no change to external api) would be a sentinel value in the `expected_origin` arg to `verify_authentication_response` - e.g. `https://*.example.com`?

---

Screenshot 2023-10-27 at 09 46 20

hugorodgerbrown:
Fixes #182 

I have added a new helper function to wrap up validation of `expected_origin` in the registration and authentication verify functions. This function handles straight str:str matches, str:list[str] matches, and a new subdomain match.

The subdomain match relies on a sentinel value of `"*"` as a wildcard to indicate that subdomains are supported. e.g. if `exepected_origin="https://*.example.com"` then the verify functions will accept "foo.example.com".

RP ID | Expected origin | Client data origin | Result
-- | -- | -- | --
example.com | https://example.com | https://example.com | match
example.com | https://*.example.com | https://foo.example.com | wildcard match
example.com | https://*.example.com | https://example.com | no wildcard match

This is the origin of the update:

> The Relying Party **MUST** validate the origin member of the client data.
>
> The Relying Party **MUST NOT** accept unexpected values of origin.

and

> Validation **MAY** be performed by exact string matching or <ins>**any other method**</ins> as needed by the Relying Party.

This PR addresses the "any other method" point by supporting a wildcard in the expected origin value. It doesn't do any structural parsing (scheme, host, domain, port) - it just does a straight str match on the prefix / suffix generated by splitting the expected origin with a wildcard character ("*").

NB This does mean that a simple "*" as the expected origin will match anything.

_From https://w3c.github.io/webauthn/#sctn-validating-origin_

### DIFF
--- a/tests/test_validate_expected_origin.py
+++ b/tests/test_validate_expected_origin.py
@@ -1,70 +1,83 @@
 from unittest import TestCase
 
 from webauthn.helpers.validate_expected_origin import (
-    match_root_domain,
+    is_exact_match,
+    is_wildcard_match,
+    match_origins,
     validate_expected_origin,
 )
 
 
 class TestValidateExpectedOrigin(TestCase):
 
-    def  test_validate_origins(self):
-        # orgin, expected_origin, result
-        test_data = [
-            # root domain match
-            ("https://example.org", ["https://example.org", "https://login.example.org"], True),
-            # subdomain match
-            ("https://login.example.org", ["https://example.org", "https://login.example.org"], True),
-            # match on non-HTTP scheme (from w3c docs)
-            ("example-os:appid:204ffa", ["https://example.org", "example-os:appid:204ffa"], True),
-            # subdomain not found in list
-            ("https://fail.example.org", ["https://example.org", "https://login.example.org"], False),
-            # subdomain fails on exact root domain main
-            ("https://fail.example.org", "https://example.org", False),
-            # subdomain passes on wildcard match
-            ("https://fail.example.org", "https://*.example.org", True),
-            # root domain failes on wildcard match
-            ("https://example.org", "https://*.example.org", False),
+    def test_is_exact_match(self):
+        """Test for case insensitive exact match."""
+        expected_origin_result = [
+            # exact match
+            ("https://www.example.org", "https://www.example.org", True),
+            # case insensitive match
+            ("https://www.EXAMPLE.org", "https://www.example.org", True),
+            # check subdomain fails on exact match
+            ("https://example.org", "https://www.example.org", False),
         ]
-        for origin, expected_origin, result in test_data:
-            self.assertEqual(
-                validate_expected_origin(expected_origin, origin),
-                result,
-                "Expected {} to match {}".format(origin, expected_origin)
-            )
+        for expected, origin, result in expected_origin_result:
+            match = is_exact_match(expected, origin)
+            self.assertEqual(match, result, "{} != {}".format(origin, expected))
 
-    def test_match_root_domain(self):
-        test_data = [
+    def test_is_wildcard_match(self):
+        """Test for http(s) wildcard match."""
+        expected_origin_result = [
             # subdomain passes on wildcard match
-            ("https://pass.example.org", "https://*.example.org", True),
+            ("https://*.example.org", "https://pass.example.org", True),
             # subdomain passes with a port
-            ("https://pass.example.org:8000", "https://*.example.org:8000", True),
+            ("https://*.example.org:8000", "https://pass.example.org:8000", True),
             # root domain fails on wildcard match
-            ("https://example.org", "https://*.example.org", False),
+            ("https://*.example.org", "https://example.org", False),
             # subdomain fails on scheme mismatch
-            ("http://fail.example.org", "https://*.example.org", False),
+            ("https://*.example.org", "http://fail.example.org", False),
             # subdomain fails on port mismatch
-            ("https://fail.example.org:8000", "https://*.example.org:8001", False),
+            ("https://*.example.org:8001", "https://fail.example.org:8000", False),
+            ("https://*.example.org", "https://fail.example.org:80", False),
+            ("https://*.example.org:80", "https://fail.example.org", False),
+            # unsupported scheme
+            ("file://path/to/*/example.org", "file://path/to/*/example.org", False),
+            # unsupported scheme
+            ("file://path/to/*/example.org", "file://path/to/*/example.org", False),
+            # not a wildcard - exact match
+            ("https://example.org", "https://example.org", True),
         ]
-        for origin, wildcard, result in test_data:
-            self.assertEqual(
-                match_root_domain(wildcard, origin),
-                result,
-                "Expected {} to match {}".format(origin, wildcard)
-            )
+        for expected, origin, result in expected_origin_result:
+            match = is_wildcard_match(expected, origin)
+            self.assertEqual(match, result, "{} != {}".format(origin, expected))
 
-    def test_match_root_domain__errors(self):
-        # check for origins that do not support wildcards
-        origins = (
-            # invalid scheme
-            "file://*.example.com",
-            # missing wildcard
-            "https://example.org",
-        )
-        for invalid_origin in origins:
-            self.assertRaises(
-                ValueError,
-                match_root_domain,
-                invalid_origin,
-                "https://foo.example.com"
-            )
+    def test_match_origins(self):
+        """Test for combined match (exact and wildcard)."""
+        expected_origin_result = [
+            # exact match
+            ("https://www.example.org", "https://www.example.org", True),
+            # wildcard match
+            ("https://*.example.org", "https://pass.example.org", True),
+            # no match
+            ("https://foo.example.org", "https://bar.example.org", False),
+        ]
+        for expected, origin, result in expected_origin_result:
+            match = match_origins(expected, origin)
+            self.assertEqual(match, result, "{} != {}".format(origin, expected))
+
+    def validate_expected_origin(self):
+        # test that validation handles a str or a list
+        expected_origin_result = [
+            # single str match
+            ("https://www.example.org", "https://www.example.org", True),
+            # list match
+            ("https://www.example.org", ["https://www.example.org"], True),
+            # list match
+            ("https://www.example.org", ["https://example.org", "https://www.example.org"], True),
+            # single str mismatch
+            ("https://www.example.org", "https://foo.example.org", False),
+            # list mismatch
+            ("https://www.example.org", ["https://foo.example.org"], False),
+        ]
+        for expected, origin, result in expected_origin_result:
+            match = validate_expected_origin(expected, origin)
+            self.assertEqual(match, result, "{} != {}".format(origin, expected))
--- a/tests/test_validate_expected_origin.py
+++ b/tests/test_validate_expected_origin.py
@@ -1,8 +1,7 @@
 from unittest import TestCase
 
+from webauthn.helpers.exceptions import InvalidExpectedOrigin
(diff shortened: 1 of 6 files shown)


webauthn/authentication/verify_authentication_response.py
```diff
@@ -102,18 +103,10 @@ def verify_authentication_response(
...
-        try:
-            expected_origin.index(client_data.origin)
-        except ValueError:
-            raise InvalidAuthenticationResponse(
-                f'Unexpected client data origin "{client_data.origin}", expected one of {expected_origin}'
-            )
+    if not validate_expected_origin(expected_origin, client_data.origin):
+        raise InvalidAuthenticationResponse(
+            f'Unexpected client data origin "{client_data.origin}", expected "{expected_origin}"'
+        )
```
hugorodgerbrown, line 109: We've lost some granularity on the error message here - but it's easy to restore if required.

MasterKale:
@hugorodgerbrown Did something happen? The work in here still seemed valid, I only objected to blocking support for localhost URLs when I closed #184. Otherwise I thought the wildcard support was a nice touch.

hugorodgerbrown:
Hi @MasterKale - apologies, I thought I'd posted a comment but apparently not.

I closed it because I noticed on re-reading that I was looking at the latest version ([Editor's Draft](https://w3c.github.io/webauthn)), which is (presumably) not published formally.

hugorodgerbrown:
> those of us in the Web Authentication Working Group

Ah - if I'd known your position on that I would have asked some more difficult question ;-)

I'll re-open - I've spent more time on the spec getting my head around the RP ID vs. expected origin checks (which was also my reservation, I wasn't totally sure I understood what was going on), so it should be tighter this time round.

I'll take a look this evening (UK time) and get it back on track.

webauthn/helpers/validate_expected_origin.py
```diff
@@ -0,0 +1,90 @@
...
+    if "*" not in expected_origin:
+        return False
+
+    try:
+        startswith, endswith = expected_origin.split("*")
+    except ValueError:
+        raise InvalidExpectedOrigin("Wildcard origin must contain exactly one '*'")
+
+    # NB this allows very broad matches, e.g. "http*" will match any web url
+    return origin.startswith(startswith) and origin.endswith(endswith)
```
hugorodgerbrown, line 63: Not sure if this is right - should we enforce some rules that prevent e.g. "*"?

tests/test_validate_expected_origin.py
```diff
@@ -0,0 +1,61 @@
...
+    validate_expected_origin,
+)
+
+
+class TestValidateExpectedOrigin(TestCase):
+
+
+    def test_match_origins(self):
+        """Test for combined match (exact and wildcard)."""
+        expected_origin_result = [
```
MasterKale, line 14: Can you please add some tests to ensure that this logic will work with `localhost` addresses too please? A few obvious local dev addresses come to mind:

- `http://localhost`
- `http://localhost:8000`
- `https://localhost` ([some devs use mkcert](https://web.dev/articles/how-to-use-local-https) to host localhost over TLS)

I _really_ don't want this to impact the (read: my 😂) ability to develop on localhost.

tests/test_validate_expected_origin.py
```diff
@@ -0,0 +1,61 @@
...
+            # exact match
+            ("https://www.example.org", "https://www.example.org", True),
+            # wildcard subdomain match
+            ("https://*.example.org", "https://pass.example.org", True),
+            # wildcard prefix-only match
+            ("example-os*", "example-os:appid:204ffa1a5af110ac483f131a1bef8a841a7a", True),
+            # wildcard suffix-only match
+            ("*.com", "https://something-random.com", True),
+            # wildcard match anything - this is a bad idea
+            ("*", "totally random origin that isn't even an origin", True),
```
MasterKale, line 24: I don't think these should be supported, the secure context that's required for WebAuthn to work requires the origin to be `http://localhost:*` or an `https://` address. Allowing use of a value like `*.com` would allow "http://example.com" which is _not_ a valid domain.

I also don't want to encourage _not_ verifying origins which is why I object to `*` as being a supported value. This should be specifically denied, I'm talking "raise an `Exception`" level of disallow.

MasterKale (changes requested):

webauthn/helpers/validate_expected_origin.py
MasterKale: Resolving this because I inadvertently replied as PR feedback below ✌️

tests/test_validate_expected_origin.py
hugorodgerbrown: I'll add an `is_secure_context` function to capture acceptable partials so we don't have such an open barn door.

hugorodgerbrown:
I'm going to suggest closing this. The commit I pushed earlier is overly complex, but highlights the fact that there is no simple answer to wildcards that prevents the "*.com" scenario that doesn't end with the question "how can I parse a subdomain", which is an SO favourite?

tl;dr you can't without maintaining a list of TLDs.

You could support a very limited subset of allowed wildcard patterns, but then if that's the case you may as well just pass in a long list of acceptable origin values, which is already supported.

It would be nice to support wildcards, but preventing abuse (/mistakes) is problematic.

MasterKale:
@hugorodgerbrown Thank you very much for attempting to pull this off. I generally agree with your conclusion, so I'm going to take you up on your suggestion to close this PR. Please don't let this dissuade you from trying again if you happen to figure out a better path forward 🙏

Stormheg:
Apologies for commenting on a closed PR, wanted to do a quick check here before opening a new issue.

The [python-fido2](https://github.com/Yubico/python-fido2) library is doing exactly what has been concluded here: **it performs a check against a hardcoded list of TLDs**. It uses a hardcoded copy from https://publicsuffix.org/

Relevant code: https://github.com/Yubico/python-fido2/blob/edcb00bc327453cfcf549ef4f12cc25a88b6b0a4/fido2/rpid.py#L44-L74

@MasterKale would you deem this an acceptable solution to implement in py_webauthn? I'm happy to submit a PR, though unsure about any licensing issues that may occur from reusing python-fido2 code.

MasterKale:
@Stormheg That's an interesting proposal, the code itself seems pretty straightforward, but **public_suffix-dat** is 311KB of plaintext with almost 7700 records...

Screenshot 2024-04-26 at 9 41 26 AM

Screenshot 2024-04-26 at 9 42 47 AM

To me that's too much "weight" to add to this library - it could make using py_webauthn very RAM-intensive, not to mention parsing and iterating through all the records on every registration because this library is not well-positioned to cache and map the parsed values in a way that every RP implementation could benefit from.

I'm willing to talk a bit more about it but right now given these facts I don't think it's a good idea to go down that path here.

Stormheg:
@MasterKale if the size of public_suffix_list.dat is a concern, removing all comments and empty lines bring it down to about 130kb. Compressing using gzip with `--best` takes it down to about 40kb.

```sh
$ cat public_suffix_list.dat.txt | grep -v ^$ | grep -v "^//" > public_suffix_list_clean.dat
$ gzip --best public_suffix_list_clean.dat
$ du -k public_suffix_list*
# size in kb...
304	public_suffix_list.dat.txt
128	public_suffix_list_clean.dat
40	public_suffix_list_clean.dat.gz
```

Packing an extra 40kb sounds reasonable to me.

Django uses a very similar mechanism to check for common passwords. The gzipped-list of passwords it uses is 80kb. Compared to python-fido2, it uses a `set` to store the parsed lines which is much faster to check if it contains a certain value. Relevant source here: https://github.com/django/django/blob/4470c2405c8dbb529501f9d78753e2aa4e9653a2/django/contrib/auth/password_validation.py#L228-L237

Granted, Django only reads and parses this file when a validating a new password. Like you mentioned, we'd be reading, decompressing, and parsing the file contents much more often without any caching. I agree this isn't desirable; but will it be a bottleneck in practice?

## Benchmark

I tried a quick benchmark and on my system; reading, decompressing, parsing and finding a needle in the gzipped list takes about 0.002 seconds on average. Doing this 10,000 times in row takes a lot longer if you can't cache results - obviously. Tested on a Macbook Pro M1 Max, Python 3.12.3

```python
import gzip
import timeit

filename = "public_suffix_list_clean.dat.gz"
needle = "appspot.com"

def read_and_check_suffix():
    with gzip.open(filename, "rt", encoding="utf-8") as f:
        suffixes = {line.strip() for line in f.readlines()}

    if needle in suffixes:
        return True
    return False

if __name__ == "__main__":
    REPEAT = 10_000
    duration = timeit.Timer(read_and_check_suffix).timeit(number=REPEAT)
    print(f"Read {REPEAT} times: {duration:.4f} seconds")
    print(f"Average per cycle: {duration / REPEAT:.5f} seconds")
```

```
$ python benchmark.py
Read 10000 times: 21.1129 seconds
Average per cycle: 0.00211 seconds
```

---

Overall, I'm not super concerned with performance issues myself. But if the answer is still no, that's acceptable for me too.

jwag956:
I took a look at the list - it is changing very frequently - so you definitely don't want to include it in this package. However possibly adding the feature to check IF the application has configured it. One way to do this is a pattern I have used In Flask-Security  - add a base-class that implements the required methods that py_webauthn would need to call (such as 'get list' or 'iterate list'. Then application can sub-class that and pass it in as part of initialization. That way - py_webauthn isn't assuming anything about how the list is maintained nor how it is stored/access by the application. And the default is 'no list'.

Stormheg:
Fair point, I had not noticed the public suffix list evolves this fast. Though [all recent changes](https://github.com/publicsuffix/list/commits/master) appear to mostly add new entries for public suffixes similar to the `*.compute.amazonaws.com` example on https://html.spec.whatwg.org/multipage/browsers.html#is-a-registrable-domain-suffix-of-or-is-equal-to.

I suppose shipping an outdated lists could pose a problem because we'd allow the origin check to pass when it should not according to the latest public suffix list. Not sure if that is enough of a con to completely scrap the idea of shipping a list. It appears to be good enough for the developers of python-fido2 and I'd feel somewhat inclined to agree with that. 100% accuracy is a lot harder to achieve, working off a semi recent list may be good enough?

Notably, python-fido2 also supports passing a custom callable parameter to override the origin check with your own if the default validation doesn't suit, see https://github.com/Yubico/python-fido2/blob/8b979313e9ac187cfa06f0882a3050f40868bcce/fido2/server.py#L147

For my own purpose, I've opted to just pass py_webauthn a list of all expected origins like what the current api needs. More configuration for me, but it is always accurate. Passing an explicit list of expected origins turns out to works well enough for me. That said, I would appreciate a way to override the check like can be done with python-fido2. That way I could implement my origin verification mechanism (and potentially shoot myself in the foot by implementing it badly? 😆)

Food for thought - would love to hear what conclusion you arrive at Matthew.

PR #186, opened 2023-11-09
Fix validation of credantials with pydantic v1

### LINKED ISSUE #185: Pydantic V1 validated AuthenticationCredential & RegistrationCredential do not pass verify
When verifying authentication or registration responses that are are "validated" pydantic models you will get an error that `id and raw_id were not equivalent`.  This is because on pydantic V1 base64 encoded bytes are not converted to bytes correctly when calling `validate()` on the model.

This is especially a problem when using this library with fastapi because it uses `validate()` internally to validate request data.
 ## Example

```python
app = fastapi()

@app.post("/passkey/verify")
def passkey_verify(authentication_response: AuthenticationCredential) -> AuthLoginResponse:
        # further logic omitted...

        # This will throw InvalidAuthenticationResponse: id and raw_id were not equivalent
        authentication = verify_authentication_response(
            credential=authentication_response,
            ...
        )
```

bschoenmaeckers:
fixes #185 and unify pydantic v1 & v2 logic.


MasterKale:
Thank you for your contribution @bschoenmaeckers. Unfortunately for this PR, I've removed Pydantic as a dependency of this project in #195 so this change is no longer needed.

PR #197, opened 2024-01-11
feat/user_id_bytes_only

### LINKED ISSUE #187: Can't use `bytes` as user ID
Given that the user handle is an `ArrayBuffer` in the WebAuthn API, I expected that the `user_id` value in this library would accept a random value of `bytes`. I was surprised to find that this is not supported.

This is probably an intentional design decision, but I thought I'd report that requiring it to be a UTF-8 caused more friction for me than allowing randomly generated bytes.

When combined with this line, I can't figure out how to generate an ID using random bytes. I guess you have to generate a random UTF-8 string instead?

If the library is going to be opinionated about adding an abstraction over the actual WebAuthn API, I think it would at least be useful to have a function to generate a random string ID.

MasterKale:
This PR only accepts `bytes` for the `user_id` argument when calling `generate_registration_options()`. Additionally, `user_id` is now **OPTIONAL**; when omitted 64 random bytes will be generated and used for this value.

Fixes #187.

### DIFF
--- a/tests/test_generate_registration_options.py
+++ b/tests/test_generate_registration_options.py
@@ -19,11 +19,12 @@ class TestGenerateRegistrationOptions(TestCase):
     @patch("secrets.token_bytes")
     def test_generates_options_with_defaults(self, token_bytes_mock: MagicMock) -> None:
         token_bytes_mock.return_value = b"12345"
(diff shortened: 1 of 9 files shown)


jwag956:
Awesome - I have my code all working with this - even easier - and thank you for now doing the b64decode as part of parse_authentication_credential_json...

PR #235, opened 2024-12-30
authentication: return user verified flag

ggirol-rc:
Some other libs like webauthn-rs enforce that when user verification is only preferred and that a device previously performed user verification, it must continue to perform it on later authentication, as it proved that it was user verification capable. Whether this out-of-spec check is desirable is an open question, however if someone wants to align behavior between a webauthn-rs-based implem and a py_webauthn implem then returning this information is necessary.

### DIFF
--- a/tests/test_verify_authentication_response.py
+++ b/tests/test_verify_authentication_response.py
@@ -43,6 +43,7 @@ class TestVerifyAuthenticationResponse(TestCase):
         assert verification.new_sign_count == 78
         assert verification.credential_backed_up == False
         assert verification.credential_device_type == "single_device"
+        assert not verification.user_verified
 
     def test_verify_authentication_response_with_RSA_public_key(self):
         credential = """{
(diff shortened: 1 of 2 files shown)


MasterKale:
Hey @ggirol-rc thanks for the suggestion. This looks fine with me, though what do you think about adding this to `verify_registration_response()` as well? I think for sake of "feature parity" across the two registration methods this should get added there too.

ggirol-rc:
Hi, I think it is already present for registration https://github.com/duo-labs/py_webauthn/blob/42537f58bb2dc073c03e7fd0f9dbf10d8c3b2c77/webauthn/registration/verify_registration_response.py#L284

MasterKale:
Oh, you're right, my bad. It's been a while since I've looked at that particular method 🤦‍♂️ 

Alright then, I'm going to kick of tests and assuming those pass I'll merge. This'll go out as part of a v2.4.1 update I'm planning.

MasterKale (approved):
lgtm :shipit:

PR #258, opened 2025-07-17
Support for PQC Added with OQS

AdityaMitra5102:
PQC Support Added with OQS. Followed draft at https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/ for COSE parameters and my draft https://datatracker.ietf.org/doc/draft-vitap-ml-dsa-webauthn/ for implementation. Tested with real responses with my under-development PQC Supported Key. The same has been added to examples as well.

Note that OQS needs to be installed. If all dependencies are there, the oqs python library automatically tries to compile OQS. If OQS is not present, it falls back to legacy,


webauthn/helpers/decoded_public_key_to_cryptography.py
```diff
@@ -61,5 +63,15 @@ def decoded_public_key_to_cryptography(
         okp_pub_key = Ed25519PublicKey.from_public_bytes(public_key.x)
 
         return okp_pub_key
+    elif isinstance(public_key, DecodedMLDSAPublicKey):
+        """
+        alg is -48, -49 (ML-DSA)
+        """
+        if public_key.alg != COSEAlgorithmIdentifier.ML_DSA_44 and public_key.alg !=COSEAlgorithmIdentifier.ML_DSA_65:
+            raise UnsupportedPublicKey(
+                f"ML-DSA Public key with unsupported algorithms"
```
MasterKale, line 72: Let's make this a bit more readable, and also report the unrecognized algorithm ID to help with troubleshooting:
```suggestion
        supported_ml_dsa_algs = (
            COSEAlgorithmIdentifier.ML_DSA_44,
            COSEAlgorithmIdentifier.ML_DSA_65,
        )
        if public_key.alg not in supported_ml_dsa_algs:
            raise UnsupportedPublicKey(
                f"ML-DSA Public key with unsupported algorithm {public_key.alg}"
```

webauthn/helpers/mldsa.py
```diff
@@ -0,0 +1,31 @@
...
+        if alg==COSEAlgorithmIdentifier.ML_DSA_65:
+            self.verifier=oqs.Signature('ML-DSA-65')
+
+    def verify(self, signature, data) -> None:
+        import oqs
+        assert self.verifier.verify(data, signature, self.pub)
+
+_ml_dsa_available=None
+
+def isML_DSA_available():
```
MasterKale, line 21:
```suggestion
def is_ml_dsa_available():
```

webauthn/registration/generate_registration_options.py
```diff
@@ -35,6 +40,10 @@ def _generate_pub_key_cred_params(
     COSEAlgorithmIdentifier.RSASSA_PKCS1_v1_5_SHA_384,
     COSEAlgorithmIdentifier.RSASSA_PKCS1_v1_5_SHA_512,
 ]
+
+if isML_DSA_available():
+    default_supported_pub_key_algs=ml_dsa_algs+default_supported_pub_key_algs
```
MasterKale, line 45:
```suggestion
if is_ml_dsa_available():
    default_supported_pub_key_algs = ml_dsa_algs + default_supported_pub_key_algs
```

webauthn/helpers/mldsa.py
```diff
@@ -0,0 +1,31 @@
+from .cose import COSEAlgorithmIdentifier
+
+class ML_DSAPublicKey:
```
MasterKale, line 3: I'm getting a little nitpicky here, but can we rename this class? I just think it looks a little weird with the underscore 🤷‍♂️ 
```suggestion
class MLDSAPublicKey:
```

webauthn/helpers/cose.py
```diff
@@ -43,6 +45,7 @@ class COSEKTY(int, Enum):
     OKP = 1
     EC2 = 2
     RSA = 3
+    MLDSA = 7
```
MasterKale, line 48: Might as well keep the naming here consistent with the way the algs above were defined:

```suggestion
    ML_DSA = 7
```

webauthn/helpers/mldsa.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, alg, pub) -> None:
+        if not isML_DSA_available():
+            raise Exception("OQS Not installed")
```
MasterKale, line 6: This should raise something new in `webauthn.helpers.exceptions` instead of a generic exception.

MasterKale (commented):
Hello @AdityaMitra5102 I apologize for the delayed review. This PR looks pretty good to me, I just had a few suggestions.

I also need to request the addition of some unit tests. It looks like you already have sample PQC responses as per the additions to the examples, but I wonder if those couldn't also become new tests added to the relevant files in **tests/**.

Based on how you try to detect the presence of `oqs`, does that mean **liboqs-python** can be considered an optional dependency of this project? I think it'd be great if, for now, py_webauthn users could continue to use the project with a simple `pip install webauthn`, then let them later opt into PQC use with `pip install liboqs-python` and then py_webauthn immediately starts using it.

MasterKale:
Hello @AdityaMitra5102 apologies for the suddenness of this but I've superseded this PR with #260 that uses Dilithium-py instead.

AdityaMitra5102:
A yes. Even I thought of using Dilithium-py at the beginning but the way the author warned about the possibilities of side channel attack, that stopped me. Tbh in my own resource constrained environments, including in stuff like circuit-python where normal libraries dont run, and even dilithium-py failed because the circuit python hashlib didnt support shake3, I ported a custom version of dilithium-py. I replaced the aes from cryptography with pyaes and hashlib for sha3 and shake with py-keccak. You might want to check it out.

It has no dependencies except numpy (which can be replaced with ulab-numpy for resource constrained devices.)

https://github.com/AdityaMitra5102/ML-DSA-PurePy

PR #268, opened 2026-05-06
Add Google's new root certificate to list of trusted certificates for key attestation

robbie-duo:
As of Feb 1, 2026 Google started using a new root cert for signing attestation certificate chains (actually closer to March 25, 2026).

Go to https://developer.android.com/privacy-and-security/security-key-attestation and click the "Root Certificates" dropdown to see this and https://android.googleapis.com/attestation/root for the current list of certs.

This PR updates the Key Attestation validation code to take the new cert into account.

Key Attestation validation is broken for all modern Andorid devices until this is merged.

### DIFF
--- a/tests/test_verify_registration_response_android_key.py
+++ b/tests/test_verify_registration_response_android_key.py
@@ -2,7 +2,18 @@ from unittest import TestCase
 from datetime import datetime
 from OpenSSL.crypto import X509Store
 
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.serialization import Encoding
+
 from webauthn.helpers import base64url_to_bytes
+from webauthn.helpers.known_root_certs import (
+    google_hardware_attestation_root_1,
+    google_hardware_attestation_root_2,
+    google_hardware_attestation_root_3,
+    google_hardware_attestation_root_4,
+    google_hardware_attestation_root_5,
+)
 from webauthn.helpers.structs import AttestationFormat
 from webauthn import verify_registration_response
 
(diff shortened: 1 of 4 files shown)


tests/test_verify_registration_response_android_key.py
```diff
@@ -60,3 +71,59 @@ def test_verify_attestation_android_key_hardware_authority(
         assert verification.credential_id == base64url_to_bytes(
             "AYNe4CBKc8H30FuAb8uaht6JbEQfbSBnS0SX7B6MFg8ofI92oR5lheRDJCgwY-JqB_QSJtezdhMbf8Wzt_La5N0"
         )
+
+
+class TestGoogleHardwareAttestationRootCerts(TestCase):
+    def test_all_root_certs_are_valid_x509(self) -> None:
```
MasterKale: I suggest, instead of any of these tests, adding a test case to the existing `TestVerifyRegistrationResponseAndroidKey` above that populates `pem_root_certs_bytes_by_fmt` with the new root cert and verifies a new response.

robbie-duo: Agreed. I was being lazy. I'll get a real registration response and will make this happen

robbie-duo: Test updated! I grabbed this info directly from a real android device against webauthn.io. I had to also update our asn1 parsing since there are newer fields we weren't handling

webauthn/helpers/asn1/android_key.py
```diff
@@ -269,6 +269,24 @@ class AuthorizationList(Sequence):
                 explicitTag=tag.Tag(tag.tagClassContext, tag.tagFormatSimple, 719)
             ),
         ),
+        OptionalNamedType(
```
robbie-duo, line 272: More fields are now added. Without this change, parsing failed at `der_decode(ext_value, asn1Spec = KeyDescription())` with a `PyAsn1Error` when it finds an unknown tag in the `AuthorizationList`

robbie-duo: Tangential to the changes I'm making here... I wonder if we can/should make this code more resilient so we don't fail parsing if there are unknown tags so if Google adds more data our lib doesn't error out

devontekaylaharrison-wq (commented):
tests/test_verify_registration_response_android_key.py

Issue #267, opened 2026-04-29
cbor2 incompatible changes (>=v6.0.0)


kramar11:
I saw that cbor2 >= 6.0.0 is a major rewrite with incompatible changes. As pyproject.toml has "cbor2>=5.6.5" this will be pulled automatically.
With cbor2 6.0.1 the "test_handles_memoryviews" fails - most likely due to those incompatible changes!

MasterKale:
Hellp @kramar11 thanks for bringing this to my attention. I was out on paternity leave when you reported this hence my delayed response. Fortunately #269 made the necessary changes to fix the compatibility issue. I'm bundling this change as part of an upcoming "v2.8.0" release. I'll follow up when the new release is available on PyPI.

MasterKale:
I'm choosing to hold off on supporing `cbor2>=6` for now because the v6 release still seems to be in flux somewhat judging by how quickly subsequent releases have come out for it. I've documented this in #272

MasterKale:
These changes are now live https://pypi.org/project/webauthn/2.8.0/ 🚀

PR #269, opened 2026-05-12
Bump cbor2 version

typestring:
Hi! This PR bumps cbor2 version to address the following CVE: https://github.com/advisories/GHSA-3c37-wwvx-h642

This is related to issue #267.

I upgraded to cbor2 v6 and fixed the failing test caused by breaking changes introduced in the new version.
also bumping to v6 requires raising the minimum supported Python version to 3.10. Since Python 3.9 is already EOL, I don’t think this should be a major issue.

If maintaining Python 3.9 support is preferred, an alternative fix would be applying this patch instead: https://github.com/typestring/py_webauthn/commit/8385a6a1c3f79f2c0a883d3c20aef092ebd5aa55

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


MasterKale (approved):
Thanks for the PR @typestring, this looks good to me. It's nice to see it's a relatively simple fix to support the latest cbor2 library 😌

And I'm fine with bumping the minimum supported version to 3.10. As per https://devguide.python.org/versions/ it's definitely time to move on from 3.9.

I'm still on parental leave so I can approve and merge but it'll take me a bit to cut a release with this fix. Maybe I can get a release out tonight 🤔

PR #277, opened 2026-06-12
Use BER encoding when parsing X.509 certificates to support older Android devices

MasterKale:
A small minority of older/cheaper Android devices (running Android <= 14) incorrectly generate `"android-key"` attestation statements. More specifically, the leaf cert in x5c is generated by on-device firmware that sets values in extensions that violates ASN.1 DER.

Recent releases of the **cryptography** Python library take a very firm stance on refusing to parse any certificates that violate DER (see in no particular order [here](https://github.com/pyca/cryptography/issues/14354), [here](https://github.com/pyca/cryptography/issues/13067), [here](https://github.com/pyca/cryptography/issues/12665)...) The parts of the code that check Android key attestations need to fall back to more lenient BER encoding to be able to support these older devices that will unlikely ever get firmware updates to fix this issue.

With that context, this PR updates certificate chain validation and `"android-key"` statement verification to use pyOpenSSL and pyasn1 instead to parse certificates and introspect extensions. This work aims to advance the spirit of "supporting more passkeys in more places."


MasterKale:
I regret that the need to support these older Android devices moves this project away from being able to use `cryptography.x509.verification`. My long-term goal has been to try and completely drop use of pyOpenSSL for cryptography, thus removing a dependency from this library, but alas, work needs to support the older devices and so this project takes on a bit more complexity.

